PR #5, opened 2021-06-04
Np2651 move user access service to Identity Service

janbj:
Focus on the CFN template, the Java code is mostly copied.


brinxmat (approved):

jitpack.yml
```diff
@@ -0,0 +1,2 @@
+jdk:
```
axthosarouris, line 1: suggetion for jitpack in order to run with the same jdk as we are using:

```
#This file runs only when one asks a version or Git hash-version from Jitpack
jdk:
  - openjdk11
before_install:
  - sdk install java 11.0.10-open
  - sdk use java 11.0.10-open
install:
  - ./gradlew clean
# Jitpack requires to copy the generated artifacts to ~/.m2 repo
  - ./gradlew build publishToMavenLocal
```

swagger.yml
```diff
@@ -0,0 +1,27 @@
+openapi: 3.0.0
```
axthosarouris, line 1: delete this file. not necessary

jitpack.yml
axthosarouris: Or if we do not need jitpack for this service maybe delete the file?

axthosarouris (changes requested):

axthosarouris (approved):

PR #14, opened 2021-06-17
Using data service from Customer Service as library

janbj:


cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.Optional;
+
+public class CustomerDbClient implements CustomerApi {
+
+    public static final String ID_NAMESPACE_ENV = "ID_NAMESPACE";
+    private static final Logger logger = LoggerFactory.getLogger(CustomerDbClient.class);
+    public static final String CUSTOMER_NOT_FOUND_FOR_ORG_NUMBER = "Customer not found for orgNumber={}";
+
+    private CustomerService service;
```
axthosarouris, line 25: both final

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
```diff
@@ -0,0 +1,74 @@
...
+    public CustomerDbClient() {
+        this(defaultCustomerService(),
+                defaultCustomerMapper()
+        );
+    }
+
+    @JacocoGenerated
+    private static DynamoDBCustomerService defaultCustomerService() {
+        return new DynamoDBCustomerService(
+                AmazonDynamoDBClientBuilder.defaultClient(),
```
axthosarouris, line 38: set Region using the AWS_REGION env variable that exists by default in all Lambdas.

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
```diff
@@ -0,0 +1,74 @@
...
+                ObjectMapperConfig.objectMapper,
+                new Environment());
+    }
+
+    @JacocoGenerated
+    private static CustomerMapper defaultCustomerMapper() {
+        String namespace = new Environment().readEnv(ID_NAMESPACE_ENV);
```
axthosarouris, line 45: Create a class with all such constants and read them before startup. Set default values that show the lack of a value  if env values are missing. e.g. 
```
NAMESPACE = ENVIRONMENT.readEnvOpt("ID_NAMESPACE").orElse("UndNamespace");
```

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public CustomerDbClient(CustomerService service, CustomerMapper mapper) {
+        this.service = service;
+        this.mapper = mapper;
+    }
+
+    @Override
+    public Optional<CustomerResponse> getCustomer(String orgNumber) {
+        CustomerResponse response = null;
```
axthosarouris, line 56:
```suggestion
@Override
    public Optional<CustomerResponse> getCustomer(String orgNumber) {
        return attempt(() -> service.getCustomerByOrgNumber(orgNumber))
                   .map(this::toCustomerResponse)
                   .toOptional(fail -> handleError(fail, orgNumber));
    }

    private void handleError(Failure<CustomerResponse> fail, String orgNumber) {
        logger.error(CUSTOMER_NOT_FOUND_FOR_ORG_NUMBER, orgNumber, fail.getException());
    }
```

cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/service/CustomerDbClientTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.UUID;
+
+import static java.lang.String.join;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class CustomerDbClientTest {
```
axthosarouris, line 22: I suppose the tests are OK for now, but should we create a task to replace these tests with a "real" service talking to a local DynamoDB?

axthosarouris (changes requested):

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
janbj: I don't agree with this. I think it is better to throw an exception if the value is missing.

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
janbj: Example?

cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/service/CustomerDbClientTest.java
janbj: Yes, we can create separate task for this.

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
axthosarouris:
```
public static final Region AWS_REGION = setupRegion();
    
    @JacocoGenerated
    private static Region setupRegion() {
        return ENVIRONMENT.readEnvOpt("AWS_REGION").map(Region::of).orElse(Region.EU_WEST_1);
    }


private static AmazonDynamoDB defaultDynamoDbClient() {
        return AmazonDynamoDBClientBuilder
                   .standard()
                   .withRegion(ApplicationConstants.AWS_REGION.id())
                   .build();
    }

janbj: Done

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/CustomerDbClient.java
janbj: Done as we agreed.

axthosarouris (approved):

brinxmat (approved):

PR #15, opened 2021-06-17
Use data service from Users and Roles Service in Cognito trigger

janbj:


cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/UserDbClient.java
```diff
@@ -0,0 +1,60 @@
...
+        return Optional.ofNullable(userDto);
+    }
+
+    @Override
+    public UserDto createUser(UserDto user) {
+        try {
+            databaseService.addUser(user);
+        } catch (InvalidEntryInternalException | ConflictException | InvalidInputException e) {
+            logger.error(ERROR_CREATING_USER, e);
+            throw new BadGatewayException(e.getMessage(), e);
```
janbj, line 46: This is a questionable error response, but was part of original contract

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/UserDbClient.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    public UserDto createUser(UserDto user) {
+        try {
+            databaseService.addUser(user);
+        } catch (Exception e) {
+            logger.error(ERROR_CREATING_USER, e);
+            throw new UserServiceException(e.getMessage(), e);
+        }
+        //TODO: maybe not necessary to return user here?
```
janbj, line 44: @axthosarouris What do you think here?

brinxmat: a) Is the resource being created in a way that entails that it exists before we return it?
b) Is it better to return the details that indicate where the resource is and an indication of when it is likely to be available?

brinxmat (approved):

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/service/UserDbClient.java
janbj: I am worried that the write to dynamodb might not be in effect at the time we read it back. If user object is not changed in any way as a part of the create operation I see no reason to return it.

cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/service/UserDbClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+        doThrow(InvalidInputException.class).when(databaseService).addUser(any());
+
+        Exception exception = Assert.assertThrows(UserServiceException.class, () -> userApi.createUser(userDto));
+
+        assertThat(exception, instanceOf(UserServiceException.class));
+        assertThat(exception.getCause(), instanceOf(InvalidInputException.class));
+    }
+
+    @Test
+    public  void updateUserNoErrorsOnSuccess() throws InvalidEntryInternalException {
```
torarnet, line 84: Double whitespace to maybe fix sometime later on

torarnet (approved):

PR #19, opened 2021-06-25
NP-2843: when a user has empty roles the role list should be an empty…

axthosarouris:
when a user has empty roles the role list should be an empty list


axthosarouris:
The failing check cannot be fixed before the new pipeline is in place

brinxmat (approved):

svenngjeto (approved):

PR #20, opened 2021-06-29
URLDecoding hosted user affiliation

svenngjeto:
AWS Cognito's property 'custom:hostedAffiliation' is read with URLDecoding before being used.

### DIFF
--- a/cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
+++ b/cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
@@ -23,6 +23,8 @@ import nva.commons.core.attempt.Try;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
```diff
@@ -249,4 +254,13 @@ private String extractAffiliation(String hostedAffiliation) {
             return EMPTY_STRING;
         }
     }
+
+    private String decodeAffiliation(String hostedAffiliation) {
+        try {
+            return URLDecoder.decode(hostedAffiliation, StandardCharsets.UTF_8);
+        } catch (IllegalArgumentException e) {
+            logger.info(PROBLEM_DECODING_HOSTED_USERS_AFFILIATION, e);
```
axthosarouris, line 262: warning better?

axthosarouris: Also maybe better to have one return point:

```
private String decodeAffiliation(String hostedAffiliation) {
        return attempt(() -> URLDecoder.decode(hostedAffiliation, StandardCharsets.UTF_8))
                   .orElse(fail -> logWarningAndReturnOriginalValue(fail, hostedAffiliation));
    }

    private String logWarningAndReturnOriginalValue(Failure<String> fail, String hostedAffiliation) {
        logger.warn(PROBLEM_DECODING_HOSTED_USERS_AFFILIATION, fail.getException());
        return hostedAffiliation;
    }
```

axthosarouris (approved):

axthosarouris (approved):

PR #29, opened 2021-09-01
Added VocabularySettings to customer

janbj:


customer-commons/src/main/java/no/unit/nva/customer/model/VocabularySetting.java
```diff
@@ -0,0 +1,94 @@
...
+                && Objects.equals(id, that.id)
+                && status == that.status;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, id, status);
+    }
+
+    public static final class Builder {
```
brinxmat, line 58: I'm growing opinionated here, I know. Not sure I would include a builder. I feel in this case particularly — where the constructed object _has_ to have all three values in order to be valid — a simple three-argument constructor is more appropriate. In general these days, I feel builders are a pattern that at best belongs in external client APIs, but I can see the point if we have god classes (in itself an issue).

customer-commons/src/main/java/no/unit/nva/customer/model/VocabularyStatus.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Lookup enum by value.
+     *
+     * @param value value
+     * @return enum
+     */
+    public static VocabularyStatus lookup(String value) {
+        return stream(values())
+                .filter(nameType -> nameType.getValue().equalsIgnoreCase(value))
+                .findAny()
```
brinxmat, line 37: Is this a use case for nva-commons SingletonCollector?

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
```diff
@@ -42,6 +44,7 @@ private CustomerDb(Builder builder) {
         setInstitutionDns(builder.institutionDns);
         setFeideOrganizationId(builder.feideOrganizationId);
         setCristinId(builder.cristinId);
+        setVocabularySettings(builder.vocabularySettings);
```
brinxmat, line 47: Two thoughts here:

  - can there be >1 defaults? Verify with PO.
  - are there other requirements on the vocab collection? (e.g. Set, ergo no duplicates — albeit unspecced? and untested; or max entries? Default initialisation? Probably more things I can't imagine off-hand)

brinxmat (changes requested):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
janbj: Good input.

customer-commons/src/main/java/no/unit/nva/customer/model/VocabularySetting.java
```diff
@@ -0,0 +1,94 @@
...
+
+        private String name;
+        private URI id;
+        private VocabularyStatus status;
+
+        private Builder() {
+        }
+
+        public static Builder builder() {
```
axthosarouris, line 67: Also, this method should be in the object class, VocabularySetting.builder()

customer-commons/src/test/java/no/unit/nva/customer/model/CustomerTest.java
```diff
@@ -30,9 +34,12 @@ public void customerMappedToJsonAndBack() throws JsonProcessingException {
     }
 
     @Test
-    public void customerMapperCanMapBetweenCustomerDtoAndCustomerDb() {
+    public void customerMapperCanMapBetweenCustomerDtoAndCustomerDb() throws JsonProcessingException {
         CustomerDb customerDb = createCustomerDb();
         CustomerDto customerDto = customerMapper.toCustomerDto(customerDb);
+
+        System.out.println(objectMapper.writeValueAsString(customerDto));
```
axthosarouris, line 41: System.out?

axthosarouris (changes requested):

customer-commons/src/test/java/no/unit/nva/customer/model/CustomerTest.java
janbj: Good catch, not supposed to be checked in.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
janbj: More than 1 default allowed, confirmed by @garshol

brinxmat (approved):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -25,6 +26,7 @@
     private String institutionDns;
     private String feideOrganizationId;
     private String cristinId;
+    private Set<VocabularySetting> vocabularySettings;
```
axthosarouris, line 29: Here we are possibly entering into a mixture of DTO and DAO models. I do not see necessarily a danger at this point as probably the changes will not be many, but any change in the DTO will automatically mean that we should migrate our database.

axthosarouris: One solution would be to create  VocabularySettingDao and `VocabularySettingDto` that are extending the VocabularySetting. Later these classes could diverge if necessary. But maybe we can leave it as it is and address the problem if and when it occurs.

axthosarouris (dismissed):

axthosarouris (approved):

brinxmat (approved):

PR #30, opened 2021-09-02
Np 3066 move more logic to separate user pool handler

axthosarouris:
Continuing my effort to separate the logic of creating the update request to Cognito in an own class. In this PR, I moved the generation of the attributes in the UserPoolEntryUpdater class

### DIFF
--- a/cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
+++ b/cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
@@ -1,10 +1,25 @@
 package no.unit.nva.cognito;
 
-import com.amazonaws.services.cognitoidp.model.AttributeType;
+import static java.util.Objects.nonNull;
+import static no.unit.nva.cognito.Constants.DYNAMODB_CLIENT;
+import static no.unit.nva.cognito.Constants.ENVIRONMENT;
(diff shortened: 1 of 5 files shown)


brinxmat (approved):

cognito-pre-token-generation/src/main/java/no/unit/nva/cognito/TriggerHandler.java
```diff
@@ -143,11 +127,14 @@ private void injectInformationForBibsysHostedCustomer(UserAttributes userAttribu
...
-        return Optional.ofNullable(userAttributes.getOrgNumber())
-                .flatMap(orgNum -> mapOrgNumberToCustomer(removeCountryPrefix(orgNum)))
-                .map(customer -> new UserDetails(userAttributes, customer))
-                .orElse(new UserDetails(userAttributes));
+    private UserDetails createUserDetails(Event event) {
+        return Optional.of(event)
+            .map(Event::getRequest)
+            .map(Request::getUserAttributes)
+            .map(UserAttributes::getOrgNumber)
+            .flatMap(orgNum -> mapOrgNumberToCustomer(removeCountryPrefix(orgNum)))
```
janbj, line 135: Minor but seeing it now maybe we should remove country prefix in its own map() call

janbj (approved):

PR #36, opened 2021-09-07
NP:3066 Revert commits regarding logging cognito info

axthosarouris:
Remove all logging used for resolving the case NP-3066.  It has been verified that when we update the custom cognito fields and we read them immediately after, we receive the updated values. Therefore the problem lies in Cognito and there is nothing we can do in the PreTokenGenerationHandler code.


svenngjeto (approved):

brinxmat (approved):

axthosarouris:
> Does it have anything to do with us using the PreTokenGeneration trigger?

No. I had added some debugging logic that exposed some information that need not to be exposed

PR #37, opened 2021-09-20
Np2581 add content negotation to customer routes

janbj:


get-customer/src/test/java/no/unit/nva/customer/get/GetCustomerHandlerTest.java
```diff
@@ -124,4 +128,29 @@ public void requestToHandlerWithMalformedIdentifierReturnsBadRequest() throws Ex
 
         assertEquals(expected, actual);
     }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void requestToHandlerWithUnsupportedAcceptHeaderReturnsUnsupportedMediaType() throws Exception {
```
janbj, line 134: Test for new content negotiation

brinxmat (approved):

user-access-handlers/src/test/java/no/unit/nva/handlers/UpdateUserHandlerTest.java
```diff
@@ -45,6 +44,7 @@
     public static final String SAMPLE_INSTITUTION = "somewhere";
     public static final String ANOTHER_ROLE = "ANOTHER_ROLE";
     public static final String SOME_OTHER_USERNAME = "SomeOtherUsername";
+    public static final String COULD_NOT_RESOLVE_SUBTYPE_OF = "Could not resolve subtype of";
```
axthosarouris, line 47: Can we move this constant to the user-access-testing library? Also, could we add a comment for this constant that this value is supplied by the external library ?

axthosarouris (changes requested):

axthosarouris (approved):

PR #40, opened 2021-09-27
Np2580 add jsonld support for customer routes

janbj:


create-customer/src/main/java/no/unit/nva/customer/create/CreateCustomerHandler.java
```diff
@@ -69,6 +73,11 @@ protected CustomerDto processInput(CustomerDto input, RequestInfo requestInfo, C
         return customerMapper.toCustomerDto(createdCustomerDb);
     }
 
+    @Override
+    protected List<MediaType> listSupportedMediaTypes() {
```
janbj, line 77: Overriding the default listSupportedMediaTypes to also support JsonLd

get-customer/src/main/java/no/unit/nva/customer/get/GetCustomerHandler.java
```diff
@@ -82,6 +85,11 @@ protected UUID getIdentifier(RequestInfo requestInfo) throws ApiGatewayException
         }
     }
 
+    @Override
+    protected List<MediaType> listSupportedMediaTypes() {
+        return List.of(MediaType.JSON_UTF_8, MediaTypes.APPLICATION_JSON_LD);
```
brinxmat, line 90: Maybe this should be a constant somewhere since we do it several times

brinxmat (changes requested):

brinxmat (approved):

axthosarouris (approved):

PR #43, opened 2021-09-28
NP-2474: tidy-up before task (merge create and get customer)

axthosarouris:
CreateCustomer and GetCustomer handlers have exactly the same dependencies and therefore it is better to put them in one package.


template.yaml
```diff
@@ -668,7 +668,7 @@ Resources:
   NvaGetCustomerFunction:
     Type: AWS::Serverless::Function
     Properties:
-      CodeUri: get-customer
+      CodeUri: create-customer
```
axthosarouris, line 671: the common package name will be create-package until I finish merging all customer handlers

PR #45, opened 2021-09-29
NP-2474: First simple version of GetVocabularyHandler

axthosarouris:
Issue: https://unit.atlassian.net/browse/NP-2474

In this PR I created a handler that returns a hardcoded response. The aim is to test it after it has been deployed and see that the handler is activated


customer/src/test/java/no/unit/nva/customer/get/GetCustomerHandlerTest.java
```diff
@@ -1,10 +1,26 @@
 package no.unit.nva.customer.get;
 
+import static no.unit.nva.customer.get.GetCustomerHandler.IDENTIFIER;
```
axthosarouris, line 3: This file had some warnings in checkstyle and I only run auto-format

customer/src/main/java/no/unit/nva/customer/get/GetControlledVocabularyHandler.java
```diff
@@ -0,0 +1,31 @@
...
+import java.net.URI;
+import no.unit.nva.customer.model.VocabularySettingDto;
+import no.unit.nva.customer.model.VocabularyStatus;
+import nva.commons.apigateway.ApiGatewayHandler;
+import nva.commons.apigateway.RequestInfo;
+
+public class GetControlledVocabularyHandler extends ApiGatewayHandler<Void, VocabularySettingDto[]> {
+
+    public static final VocabularySettingDto HARDCODED_RESPONSE = new VocabularySettingDto(
+        "sampleVocabulary",
```
brinxmat, line 14: Do I need to say…?

brinxmat (approved):

PR #46, opened 2021-09-29
Np2589 get all customers jsonld response

janbj:


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -175,34 +39,26 @@ public boolean equals(Object o) {
...
-            && Objects.equals(getDisplayName(), that.getDisplayName())
-            && Objects.equals(getShortName(), that.getShortName())
-            && Objects.equals(getArchiveName(), that.getArchiveName())
-            && Objects.equals(getCname(), that.getCname())
-            && Objects.equals(getInstitutionDns(), that.getInstitutionDns())
-            && Objects.equals(getFeideOrganizationId(), that.getFeideOrganizationId())
-            && Objects.equals(getCristinId(), that.getCristinId())
-            && Objects.equals(getVocabularySettings(), that.getVocabularySettings())
-            && Objects.equals(getContext(), that.getContext());
+                && Objects.equals(getIdentifier(), that.getIdentifier())
```
brinxmat, line 42: Er dette riktig? Tipper at regen gjør noe her…kan ta feil

brinxmat (changes requested):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
janbj: Hva tenker du på?

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerList.java
```diff
@@ -18,34 +22,25 @@ public CustomerList() {
 
     }
 
-    public CustomerList(List<CustomerDto> customers) {
+    public CustomerList(URI id, List<CustomerDtoWithoutContext> customers) {
```
axthosarouris, line 25: I think it would be simpler if CustomerList accepted CustomerDtos.
```
    public CustomerList(List<CustomerDto> customers) {
        this.id = NAMESPACE;
        this.customers = extractCustomers(customers);
        this.context = CustomerMapper.CONTEXT;
    }
    private List<CustomerDtoWithoutContext> extractCustomers(List<CustomerDto> customers) {
        return Optional.ofNullable(customers)
            .stream()
            .filter(Objects::nonNull)
            .flatMap(Collection::stream)
            .filter(Objects::nonNull)
            .map(CustomerDto::withoutContext)
            .collect(Collectors.toList());
    }
```

axthosarouris: There is a problem with a test expecting a non empty list containing a null object.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerMapper.java
```diff
@@ -17,6 +24,22 @@
     private CustomerMapper() {
     }
 
+    private static CustomerDtoWithoutContext toCustomerDtoWithoutContext(CustomerDto customerDto) {
```
axthosarouris, line 27: this belongs to object CustomerDto. 
```
 
    public CustomerDtoWithoutContext withoutContext() {
        return attempt(() -> JsonUtils.objectMapper.convertValue(this, CustomerDtoWithoutContext.class)).orElseThrow();
    }
```

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerMapper.java
```diff
@@ -17,6 +24,22 @@
...
+        return objectMapper.convertValue(customerDto, CustomerDtoWithoutContext.class);
+    }
+
+    public static CustomerList toCustomerListFromCustomerDtosWithoutContexts(
+            List<CustomerDtoWithoutContext> customerDtos) {
+        URI id = URI.create(getIdNamespace());
+        return new CustomerList(id, customerDtos);
```
axthosarouris, line 34: id should not be passed as a parameter. NAMESPACE is a public static variable of the system.

customer-commons/src/test/java/no/unit/nva/customer/model/CustomerTest.java
```diff
@@ -47,6 +49,22 @@ public void customerMapperCanMapBetweenCustomerDtoAndCustomerDb() {
...
+        assertNotNull(customerList);
+    }
+
+    @Test
+    public void customerMapperCanMapCustomerDbToCustomerDto() {
+        CustomerDb customerDb = createCustomerDb();
+        CustomerDto customerDto = customerDb.toCustomerDto();
+        assertNotNull(customerDto);
+        assertNotNull(customerDto.getContext());
+    }
```
axthosarouris, line 66: this test is covered in CustomerDbTest. Just add a the test that the context is not null

axthosarouris (changes requested):

axthosarouris:
I do not understand this test:
```
    @Test
    public void customerListFromNull() {
        List<CustomerDto> list = new ArrayList<>();
        list.add(null);
        CustomerList customerList = new CustomerList(list);
        assertEquals(1, customerList.getCustomers().size());
        assertTrue(customerList.getCustomers().get(0) == null);
    }
```
Why do we want a non empty list with a null object instead of an empyt list.

brinxmat (dismissed):

axthosarouris (approved):

PR #47, opened 2021-09-29
NP-2474: Service returns DTOs 

axthosarouris:

PR #48, opened 2021-09-29
NP-2474: create logic for reliable converting of DTOs to DAOs and vive versa

axthosarouris:
CustomerService exposes the DAOs to the rest of the platform and that makes things quite complicated.
At the same time CustomerMapper is not really a statefull class.

This is the first step of refactoring the customer service  where the final goal is the following:
1. The DAOs will contain the logic to be converted to DTOs and vice versa.
2. The DAOs will be as short-lived as possible, i.e., the customer service will only communicate with its clients through the DTOs. That is the handlers do not need to know about the DAOs.
3. The CustomerMapper can be a lot thinner.  

In this PR we create logic to map the DAOs to DTOs 

This  PR is part of the PR #47

### DIFF
--- a/buildSrc/src/main/groovy/nva.identity.service.java-common-conventions.gradle
+++ b/buildSrc/src/main/groovy/nva.identity.service.java-common-conventions.gradle
@@ -28,11 +28,11 @@ repositories {
 
 project.ext {
     nvaCommonsVersion = '1.12.0'
-    awsSdkVersion ="1.12.59"
-    jacksonVersion ="2.10.3"
+    awsSdkVersion = "1.12.59"
+    jacksonVersion = "2.10.3"
     xraySdkVersion = '2.9.1'
     DynamoDBLocalVersion = '1.15.0'
-    nvaTestUtilsVersion = '0.1.13'
+    nvaTestUtilsVersion = '0.1.24'
     zalandoProblemVersion = '0.25.0'
     awsLambdaCoreVersion = '1.2.1'
     jupiterVersion = "5.6.0"
@@ -78,6 +78,8 @@ dependencies {
     implementation group: 'com.amazonaws', name: 'aws-xray-recorder-sdk-aws-sdk'
     implementation group: 'com.amazonaws', name: 'aws-xray-recorder-sdk-aws-sdk-instrumentor'
 
+    testImplementation group: 'org.javers', name: 'javers-core', version: '6.2.5'
+    testImplementation group: 'com.github.javafaker', name: 'javafaker', version: '1.0.2'
     testImplementation group: 'com.github.BIBSYSDEV', name: 'nva-testutils', version: project.ext.nvaTestUtilsVersion
     testImplementation group: 'com.amazonaws', name: 'DynamoDBLocal', version: project.ext.DynamoDBLocalVersion
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '3.3.3'
@@ -105,10 +107,12 @@ tasks.named('test') {
     testLogging {
         events('skipped', 'passed', 'failed')
     }
-    final Map<String, String> envVariables = new HashMap<>()
-    envVariables.put("ALLOWED_ORIGIN", "*")
-    environment(envVariables)
     finalizedBy jacocoTestReport
+}
+
+test {
+    environment "ID_NAMESPACE", "https://www.example.org"
+    environment "ALLOWED_ORIGIN", "*"
 }
 
 pmd {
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/groovy/nva.identity.service.java-common-conventions.gradle
```diff
@@ -105,12 +107,14 @@ tasks.named('test') {
...
         events('skipped', 'passed', 'failed')
     }
-    final Map<String, String> envVariables = new HashMap<>()
-    envVariables.put("ALLOWED_ORIGIN", "*")
-    environment(envVariables)
     finalizedBy jacocoTestReport
 }
 
+test {
+    environment "ID_NAMESPACE", "https://www.example.org"
```
axthosarouris, line 114: ENV variables used during testing

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
```diff
@@ -32,19 +34,24 @@
...
-        setCname(builder.cname);
-        setInstitutionDns(builder.institutionDns);
-        setFeideOrganizationId(builder.feideOrganizationId);
-        setCristinId(builder.cristinId);
-        setVocabularySettings(builder.vocabularySettings);
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static CustomerDb fromCustomerDto(CustomerDto dto) {
```
axthosarouris, line 41: explicit conversion of a DTO to a DAO. Conversion is tested to verify that it contains all fields.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
```diff
@@ -197,86 +204,110 @@ public void setVocabularySettings(Set<VocabularySettingDb> vocabularySettings) {
...
-        private String displayName;
-        private String shortName;
-        private String archiveName;
-        private String cname;
-        private String institutionDns;
-        private String feideOrganizationId;
-        private String cristinId;
-        private Set<VocabularySettingDb> vocabularySettings;
+
+        private final CustomerDb customerDb;
```
axthosarouris, line 245: Using an object directly allows us to define the logic of setting values (e.g. setVocabularySettings) only once and have better control of the set values.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerMapper.java
```diff
@@ -1,44 +1,60 @@
...
+    public static final String ID_NAMESPACE_ENV = "ID_NAMESPACE";
+    public static final String NAMESPACE = getIdNamespace();
     public static final URI NO_CONTEXT = null;
-    public static final URI context = URI.create("https://bibsysdev.github.io/src/customer-context.json");
+    public static final URI CONTEXT = URI.create("https://bibsysdev.github.io/src/customer-context.json");
     private static final ObjectMapper objectMapper = JsonUtils.objectMapper;
 
-    private final String namespace;
-
+    @SuppressWarnings("PMD.UnusedFormalParameter")
```
axthosarouris, line 21: Preparation for class to become a much thinner utility class

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerMapper.java
```diff
@@ -1,44 +1,60 @@
...
     public CustomerMapper(String namespace) {
-        this.namespace = namespace;
+
+    }
+
+    public static CustomerDto addContext(CustomerDto customerDto) {
```
axthosarouris, line 26: this is the replacement for `toCustomerDto`

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
brinxmat: Are we to understand CustomerDb as CustomerDao? I guess renaming would make this clearer

customer-commons/src/test/java/no/unit/nva/customer/model/CustomerDbTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+class CustomerDbTest {
+
+    public static final Faker FAKER = Faker.instance();
+    public static final Random RANDOM = new Random(System.currentTimeMillis());
+    public static final Javers JAVERS = JaversBuilder.javers().build();
+
+    @Test
+    public void toCustomerDtoReturnsDtoWithoutLossOfInformation() {
+        CustomerDb expected = crateSampleCustomerDb();
```
brinxmat, line 30: crate -> create

brinxmat (changes requested):
Not commenting too hard on this as it is a stepping stone

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDb.java
janbj: I support renaming it to CustomerDao

axthosarouris: I agree, but I want to do it as own PR.

janbj (approved):

PR #52, opened 2021-09-30
Refactoring cleanup

axthosarouris:


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerMapper.java
```diff
@@ -1,35 +0,0 @@
-package no.unit.nva.customer.model;
```
axthosarouris, line 1: renamed to LinkedDataContextUtils

customer-commons/src/main/java/no/unit/nva/customer/model/LinkedDataContextUtils.java
```diff
@@ -0,0 +1,31 @@
...
+        return Optional.ofNullable(customerDto)
+            .map(CustomerDto::copy)
+            .map(copy -> copy.withContext(LINKED_DATA_CONTEXT_VALUE))
+            .map(copy -> copy.withId(toId(customerDto.getIdentifier())))
+            .map(Builder::build)
+            .orElse(null);
+    }
+
+    public static URI toId(UUID identifier) {
+        return new UriWrapper(ID_NAMESPACE).addChild(identifier.toString()).getUri();
```
axthosarouris, line 29: UriWrapper allows URI path manipulation without having to think whether the parent URI ends with a slash or not

janbj (approved):

brinxmat (approved):

PR #54, opened 2021-10-01
NP-2474: getVocabularies returns vocabullary settings of user

axthosarouris:


customer-commons/build.gradle
```diff
@@ -16,9 +16,16 @@ dependencies {
 
 
 
-
+    testImplementation project (":customer-testing")
```
axthosarouris, line 19: moved the DynamoDB mocking in a separate module for resusability.

customer-commons/build.gradle
```diff
@@ -16,9 +16,16 @@ dependencies {
...
     testImplementation (group: 'com.amazonaws', name: 'DynamoDBLocal', version: project.ext.DynamoDBLocalVersion) {
         exclude group: 'com.amazonaws', module: 'aws-java-sdk-dynamodb'
         exclude group: 'com.amazonaws', module: 'aws-java-sdk-core'
     }
+}
+
+
+test{
+    environment "TABLE_NAME","nva_customers"
```
axthosarouris, line 28: there is no reason for the indices to be set by ENV variables. The nature of the indices (and therefore their name) depends on their use by the code. Changing their name has no use. In next PR I will clean this up.

customer-commons/src/test/java/no/unit/nva/customer/service/impl/DynamoDBCustomerServiceTest.java
```diff
@@ -19,65 +17,51 @@
...
-import org.junit.jupiter.migrationsupport.rules.EnableRuleMigrationSupport;
 
-@EnableRuleMigrationSupport
-public class DynamoDBCustomerServiceTest {
-
-    public static final String NVA_CUSTOMERS_TEST = "nva_customers_test";
-
-    @Rule
-    public CustomerDynamoDBLocal db = new CustomerDynamoDBLocal();
+public class DynamoDBCustomerServiceTest extends CustomerDynamoDBLocal {
```
axthosarouris, line 41: removed Junit4 rule

customer-testing/build.gradle
```diff
@@ -3,5 +3,15 @@ plugins {
 }
 
 dependencies {
-    implementation group: 'com.github.bibsysdev', name: 'apigateway'
+    implementation(group: 'com.amazonaws', name: 'DynamoDBLocal', version: project.ext.DynamoDBLocalVersion) {
```
axthosarouris, line 6: customer-testing contains only database and testing libraries.

customer/build.gradle
```diff
@@ -15,5 +15,12 @@ dependencies {
     implementation group: 'com.amazonaws', name: 'aws-lambda-java-core'
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-lambda'
     implementation group: 'com.amazonaws', name: 'aws-java-sdk-dynamodb'
+
+    testImplementation project(":customer-testing")
 }
 
+test{
+    environment "TABLE_NAME","nva_customers"
+    environment "BY_ORG_NUMBER_INDEX_NAME", "byOrgNumber"
```
axthosarouris, line 24: replication that will be cleanup in later PR

customer/src/test/java/no/unit/nva/customer/testing/CustomerDataGenerator.java
```diff
@@ -0,0 +1,106 @@
+package no.unit.nva.customer.testing;
```
axthosarouris, line 1: Code copied from customer-commons, because we have no access to the test classes. Cannot be put to customer-testing due to circular dependencies if we do so. (customer-commons already depends on customer-testing). Possible solution would be to move the datamodel in its own module

brinxmat (approved):

PR #56, opened 2021-10-01
NP-2474:remove region

axthosarouris:


customer/src/main/java/no/unit/nva/customer/get/GetControlledVocabularyHandler.java
```diff
@@ -73,7 +72,6 @@ private static AmazonDynamoDB defaultDynamoDbClient() {
             new EndpointConfiguration(Constants.AWD_DYNAMODB_SERVICE_END_POINT, Constants.AWS_REGION);
         return AmazonDynamoDBClientBuilder.standard()
             .withCredentials(new DefaultAWSCredentialsProviderChain())
-            .withRegion(AWS_REGION)
```
axthosarouris, line 76: one region on endoint should be configured

brinxmat (approved):

PR #58, opened 2021-10-05
Np 3212 post vocabulary

axthosarouris:
Handler for creating vocabulary settings for user. I suggest reading the tests first.


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -12,7 +11,6 @@
 import no.unit.nva.customer.model.interfaces.Customer;
 import nva.commons.core.JacocoGenerated;
 
-@JsonTypeName("Customer")
```
axthosarouris, line 15: JsonType is defined in the interface

brinxmat (approved):

PR #64, opened 2021-10-05
 rename VocabulaySetting to Vocabulary

axthosarouris:
Changed the names of the types and classes to fit with the specification as defined in  https://gist.github.com/brinxmat/ce87ac0a739b2060fcd8091bab857175
Changes:
1. Renamed VocabularySetting(s) to Vocabulary(ies)
2. Added get/setVocabulary to Customer interface for consistency


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -19,7 +20,8 @@
     property = "type",
     defaultImpl = CustomerDao.class
 )
-public class CustomerDao implements Customer {
+@JsonTypeName("Customer")
```
axthosarouris, line 23: added type of CustomerDao explicitly

brinxmat (approved):

PR #66, opened 2021-10-06
Fix vocabulary list

axthosarouris:


buildSrc/src/main/groovy/nva.identity.service.java-common-conventions.gradle
```diff
@@ -40,7 +40,7 @@ tasks.named('test') {
 }
 
 test {
-    environment "ID_NAMESPACE", "https://www.example.org"
+    environment "ID_NAMESPACE", "https://www.example.org/namespace"
```
brinxmat, line 43: Technically, I'd assume an NS has a trailing / or #

brinxmat (changes requested):

brinxmat (approved):

PR #71, opened 2021-10-12
return empty collections instead of nulls

axthosarouris:


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -185,12 +186,12 @@ public int hashCode() {
 
     @Override
     public Set<VocabularyDao> getVocabularies() {
-        return vocabularies;
+        return nonNull(vocabularies)? vocabularies:Collections.emptySet();
```
brinxmat, line 189: Spacing

brinxmat (approved):

PR #73, opened 2021-10-19
Np 3014 update curator view handler

axthosarouris:
Implmented the first part of [NP 3014](https://unit.atlassian.net/browse/NP-3014). I have created an UpdateViewingScope that accepts a ViewingScope and adds it in the User entry. There is no "create curator scope" because we are only updating a user profile and also the requirement is that once a user is curator they will get a default scope automatically.

### DIFF
--- a/user-access-internal-model/src/test/java/no/unit/nva/useraccessmanagement/dao/UserDbTest.java
+++ b/user-access-internal-model/src/test/java/no/unit/nva/useraccessmanagement/dao/UserDbTest.java
@@ -156,7 +156,7 @@ public class UserDbTest {
     }
 
     @Test
-    void copyShouldReturnBuilderWithFilledInFields() throws InvalidEntryInternalException, BadRequestException {
+    void shouldReturnCopyWithFilledInFields() throws InvalidEntryInternalException, BadRequestException {
         UserDb originalUser = randomUserDb();
         UserDb copy = originalUser.copy().build();
         assertThat(copy, is(equalTo(originalUser)));


cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
```diff
@@ -314,7 +314,7 @@ private String compareAssignedRoles(List<String> rolesBeforeLogin, List<String>
 
     private UserDto createUserWithCustomRole(String manuallyAssignedRole) throws InvalidEntryInternalException {
         UserDto sampleUser = createUserWithInstitutionAndCreatorRole();
-        List<RoleDto> roles = sampleUser.getRoles();
+        Set<RoleDto> roles = sampleUser.getRoles();
```
axthosarouris, line 317: User roles become a set because order is not always kept in tests and order is not important

user-access-commons/src/main/java/no/unit/nva/useraccessmanagement/model/ViewingScope.java
```diff
@@ -8,31 +8,33 @@
...
+    public static final String VIEWING_SCOPE_TYPE = "ViewingScope";
     @JsonProperty(INCLUDED_UNITS)
     private final Set<URI> includedUnits;
     @JsonProperty(EXCLUDED_UNIS)
     private final Set<URI> excludedUnits;
 
     @JsonCreator
     public ViewingScope(@JsonProperty(INCLUDED_UNITS) Set<URI> includedUnits,
-                        @JsonProperty(EXCLUDED_UNIS) Set<URI> excludedUnits) {
+                        @JsonProperty(EXCLUDED_UNIS) Set<URI> excludedUnits) throws BadRequestException {
```
axthosarouris, line 34: viewing scope throws now BadRequestException. This created quite many updates in method signatures.

user-access-internal-model/src/test/java/no/unit/nva/useraccessmanagement/dao/UserDbTest.java
```diff
@@ -135,20 +156,27 @@ void serializationThrowsExceptionWhenUserHasNullUserName() {
...
     @Test
-    void copyShouldReturnBuilderWithFilledInFields() throws InvalidEntryInternalException {
-        UserDb originalUser = UserDb.newBuilder()
-            .withUsername(SOME_USERNAME)
-            .withGivenName(SOME_GIVEN_NAME)
-            .withFamilyName(SOME_FAMILY_NAME)
-            .withInstitution(SOME_INSTITUTION)
-            .withRoles(SAMPLE_ROLES)
-            .build();
+    void copyShouldReturnBuilderWithFilledInFields() throws InvalidEntryInternalException, BadRequestException {
```
brinxmat, line 159: Test name here is shouldReturnNewObject?

user-access-handlers/src/test/java/no/unit/nva/handlers/UpdateViewingScopeHandlerTest.java
```diff
@@ -0,0 +1,143 @@
...
+        throws IOException, InvalidInputException, ConflictException {
+        var sampleUser = addSampleUserToDb();
+        var request = createInvalidUpdateViewingScopeRequest(sampleUser);
+        handler.handleRequest(request, output, CONTEXT);
+        GatewayResponse<Void> response = GatewayResponse.fromOutputStream(output);
+        assertThat(response.getStatusCode(), is(equalTo(HttpURLConnection.HTTP_BAD_REQUEST)));
+    }
+
+    @Test
+    void shouldReturnProblemWhenRequestIsNotSuccessful()
```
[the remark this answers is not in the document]
janbj: Yes, both the NotFound and BadRequest above should probably be Problem not Void.

PR #88, opened 2022-01-06
fix dependencies and use one dynamo client in trigger function

axthosarouris:
Changes:
1. Use one dynamo client in Trigger function for UserService and CustomerService
2. Fix some dependencies.


cognito-pre-token-generation/build.gradle
```diff
@@ -12,10 +12,9 @@ dependencies {
     implementation project(':customer-commons')
 
     implementation libs.nva.apigateway
-    implementation libs.bundles.jackson
+//    implementation libs.bundles.jackson
```
axthosarouris, line 15: will be removed in next PR if identity service continues working

brinxmat (approved):

PR #92, opened 2022-01-11
NP-3015:validate ViewingScope URIs

axthosarouris:
Validation for URIs submitted to ViewingScope. For now, the only check is that they should be URIs that start with 
`https://api.{env}.nva.aws.unit.no/cristin`

### DIFF
--- a/cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
+++ b/cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
@@ -444,9 +444,8 @@ public class TriggerHandlerTest {
         return userWithInstitution(userWithRoles(roles));
     }
 
-    private UserDto userWithInstitution(UserDto user) throws InvalidEntryInternalException, BadRequestException {
-
-        ViewingScope viewingScope = new ViewingScope(Set.of(SAMPLE_CRISTIN_ORG_ID), null);
+    private UserDto userWithInstitution(UserDto user) throws InvalidEntryInternalException {
+        ViewingScope viewingScope =  ViewingScope.defaultViewingScope(SAMPLE_CRISTIN_ORG_ID);
         return user.copy().withInstitution(SAMPLE_CUSTOMER_ID).withViewingScope(viewingScope).build();
     }
 
--- a/cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
+++ b/cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
@@ -1,44 +1,6 @@
 package no.unit.nva.cognito;
 
-import com.amazonaws.services.cognitoidp.AWSCognitoIdentityProvider;
-import com.amazonaws.services.cognitoidp.model.AdminGetUserRequest;
-import com.amazonaws.services.cognitoidp.model.AdminGetUserResult;
-import com.amazonaws.services.cognitoidp.model.AdminUpdateUserAttributesRequest;
-import com.amazonaws.services.cognitoidp.model.AdminUpdateUserAttributesResult;
-import com.amazonaws.services.cognitoidp.model.AttributeType;
-import com.amazonaws.services.lambda.runtime.Context;
-import no.unit.nva.cognito.model.CustomerResponse;
-import no.unit.nva.cognito.model.Event;
-import no.unit.nva.cognito.model.Request;
-import no.unit.nva.cognito.model.UserAttributes;
-import no.unit.nva.cognito.service.CustomerApi;
-import no.unit.nva.cognito.service.UserApiMock;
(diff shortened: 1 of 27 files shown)


cognito-pre-token-generation/src/test/java/no/unit/nva/cognito/TriggerHandlerTest.java
```diff
@@ -81,10 +81,12 @@
 
     public static final String CREATOR = "Creator";
     public static final String USER = "User";
-    public static final String SAMPLE_CRISTIN_ID = "http://cristin.id";
+
     public static final AdminUpdateUserAttributesResult UNUSED_RESULT = null;
     public static final Javers JAVERS = JaversBuilder.javers().build();
     public static final Context CONTEXT = mock(Context.class);
+    public static final URI SAMPLE_CRISTIN_ORG_ID = randomCristinOrgId();
```
axthosarouris, line 88: Create a valid CristinId instead of an arbitrary one. A valid Cristin ID is a  URI that has the same host as the identity service and the path starts with /cristin/organization

customer-testing/src/main/java/no/unit/nva/customer/testing/CustomerDataGenerator.java
```diff
@@ -96,6 +99,13 @@ public static UUID randomIdentifier() {
         return UUID.randomUUID();
     }
 
+    public static URI randomCristinOrgId() {
```
axthosarouris, line 102: this method is written in two places, once for the customer service and once for the user service. These services do not share any logic.

user-access-commons/src/main/java/no/unit/nva/useraccessmanagement/model/ViewingScope.java
```diff
@@ -83,26 +82,39 @@ public boolean equals(Object o) {
             return false;
         }
         ViewingScope that = (ViewingScope) o;
-        return isRecursive() == that.isRecursive()
```
axthosarouris, line 86: removed recursion because it is not used now, and it made refactoring much harder.

user-access-commons/src/main/java/no/unit/nva/useraccessmanagement/model/ViewingScope.java
```diff
@@ -83,26 +82,39 @@ public boolean equals(Object o) {
...
         if (includedUnits.isEmpty()) {
             throw new BadRequestException("Invalid Viewing Scope: \"includedUnits\" cannot be empty");
         }
+        validate(includedUnits);
+        validate(excludedUnits);
+    }
+
+    private void validate(Set<URI> uris) throws BadRequestException {
+        for (URI uri : uris) {
+            validate(uri);
```
axthosarouris, line 95: not using stream because of the checked Exception.

user-access-internal-model/src/test/java/no/unit/nva/useraccessmanagement/dao/UserDbTest.java
```diff
@@ -248,19 +249,6 @@ void shouldContainListOfCristinOrganizationIdsThatDefineCuratorsScope()
                    containsInAnyOrder(someCristinUnit, someOtherCristinUnit));
     }
 
-    @Test
-    @DisplayName("should contain field that informs whether the viewing scope should include the children of the "
-                 + "included organizations ")
-    void shouldContainFieldInformingIfTheViewingScopeIncludesTheChildrenOfTheIncludedOrgs() throws BadRequestException {
```
axthosarouris, line 254: this functionality has been removed, because we are not using it at the moment

user-access-public-model/build.gradle
```diff
@@ -1,5 +1,6 @@
 plugins {
     id 'nva.identity.service.java-library-conventions'
+    id 'nva.identity.service.dynamodblocal'
```
axthosarouris, line 3: added the plugin because the testing package depends on the dynamodb-local package

template.yaml
```diff
@@ -68,6 +68,9 @@ Globals:
     Timeout: 10
     Runtime: java11
     Tracing: Active
+    Environment:
```
axthosarouris, line 71: Set the API_HOST env variable for all lambdas

user-access-commons/src/main/java/no/unit/nva/useraccessmanagement/model/ViewingScope.java
```diff
@@ -83,26 +78,40 @@ public boolean equals(Object o) {
...
+        }
+        return null;
+    }
+
+    private static boolean pathIsExpectedPath(URI uri) {
+        return uri.getPath().startsWith(ServiceConstants.CRISTIN_PATH);
+    }
+
+    private static boolean hostIsExpectedHost(URI uri) {
+        return ServiceConstants.API_HOST.equals(uri.getHost());
```
brinxmat, line 115: are these ever used without. negation?

brinxmat (dismissed):

brinxmat (approved):

PR #130, opened 2022-01-25
NP-3544: type of customerId reflects type of id stored in database.

axthosarouris:


user-access-event-handlers/src/test/java/no/unit/nva/useraccess/events/service/UserMigrationServiceTest.java
```diff
@@ -187,8 +188,11 @@ private UserDto createSampleUser() {
                 .withUsername(randomString())
                 .withGivenName(randomString())
                 .withFamilyName(randomString())
-                .withInstitution(randomUUID().toString())
+                .withInstitution(randomInstitutionUri())
                 .build();
     }
 
+    private URI randomInstitutionUri() {
+        return URI.create("https://www.example.com/" + UUID.randomUUID());
```
axthosarouris, line 196: migration service extracts the identifier in order to send the query to customer service

PR #142, opened 2022-01-31
Removing only orgunitids pointing to cristin api

janbj:


brinxmat (approved):

user-access-event-handlers/src/main/java/no/unit/nva/useraccess/events/service/UserMigrationServiceImpl.java
```diff
@@ -47,11 +51,15 @@ private void removeOldPatternOrganizationIds(String username) {
             var systemControlNumber = authority.get().getSystemControlNumber();
             var organizationIds = authority.get().getOrganizationIds();
             organizationIds.stream()
-                .filter(ViewingScope::isNotValidOrganizationId)
+                .filter(this::isUriToCristinApi)
                 .forEach(uri -> deleteFromAuthority(systemControlNumber, uri));
         }
     }
 
+    private boolean isUriToCristinApi(URI organizationId) {
```
axthosarouris, line 59: private boolean isUriToCristinApi(URI organizationId) {
        return CRISTIN_API_URI.equals(organizationId.getHost());
    }

user-access-event-handlers/src/main/java/no/unit/nva/useraccess/events/service/UserMigrationServiceImpl.java
```diff
@@ -17,6 +17,8 @@
 
 public class UserMigrationServiceImpl implements UserMigrationService {
 
+    public static final String CRISTIN_API_URI = "https://api.cristin.no";
```
axthosarouris, line 20: will this work for dev,sandbox,test?

user-access-event-handlers/src/test/java/no/unit/nva/useraccess/events/service/UserMigrationServiceTest.java
```diff
@@ -42,6 +43,7 @@
 class UserMigrationServiceTest {
 
     private static final URI SAMPLE_ORG_ID = URI.create("https://localhost/cristin/organization/123.0.0.0");
+    private static final URI SAMPLE_CRISTIN_API_ORG_ID = URI.create(CRISTIN_API_URI + randomString());
```
axthosarouris, line 46: if you use the host variable instead of URI this has to change I suppose to something like 

```
private static final URI SAMPLE_CRISTIN_API_ORG_ID =
        new UriWrapper("https",CRISTIN_API_URI).addChild(randomString()).getUri();
```

user-access-event-handlers/src/main/java/no/unit/nva/useraccess/events/service/UserMigrationServiceImpl.java
axthosarouris: I suspect we will need this to be an env variable and therefore I find it much easier if we give only the hosts, for all external services and not URIs. If we give URIs, then we have to check, is it https, is it this, is it that.

PR #147, opened 2022-02-04
Removig old pattern org ids was never dependent on knowing org id on …

janbj:
…user


user-access-event-handlers/src/main/java/no/unit/nva/useraccess/events/service/UserMigrationServiceImpl.java
```diff
@@ -31,18 +31,21 @@ public UserMigrationServiceImpl(CustomerService customerService, BareProxyClient
     @Override
     public UserDto migrateUser(UserDto user) {
         logger.trace("Updating user:{}", user.getUsername());
+        resetViewingScope(user);
+        removeOldPatternOrganizationIds(user.getUsername());
```
janbj, line 35: This should be run independent on user having a customer set

axthosarouris (dismissed):

brinxmat (approved):

PR #155, opened 2022-04-11
Refactoring for better readability

axthosarouris:


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -298,7 +298,10 @@ private UserDto createNewUserObject(CustomerDto customer,
         return user.build();
     }
 
-    private String formatUsername(CristinPersonResponse cristinResponse, CustomerDto customer) {
+    // Create a username that will allow the user to access their resources even if the identity service stack
+    // gets totally destroyed.
+    private String createConsistentUsernameBasedOnPersonIdentifierAndOrgIdentifier(CristinPersonResponse cristinResponse,
```
axthosarouris, line 303: I think this is an important renaming, explaining why I have created the usernames in this way

brinxmat (approved):

PR #163, opened 2022-04-12
update requirements for the case a user has wrong customer id

axthosarouris:


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/AuthenticationInformation.java
```diff
@@ -157,6 +147,26 @@ private static String extractFeideIdentifier(Map<String, String> userAttributes)
         return Optional.ofNullable(userAttributes.get(FEIDE_ID)).orElse(null);
     }
 
+    private Stream<URI> allAffiliationsWithSameParentInstitution(URI parentInstitution) {
+        return this.personAffiliations.stream()
+            .filter(pair -> pair.getParentInstitution().equals(parentInstitution))
```
brinxmat, line 152: pair?

brinxmat (approved):

PR #165, opened 2022-04-21
Move user-creation logic to own class

axthosarouris:


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserEntriesCreatorForPerson.java
```diff
@@ -0,0 +1,184 @@
...
+import no.unit.nva.cognito.cristin.person.CristinPersonResponse;
+import no.unit.nva.customer.model.CustomerDto;
+import no.unit.nva.customer.service.CustomerService;
+import no.unit.nva.database.IdentityService;
+import no.unit.nva.useraccessservice.model.RoleDto;
+import no.unit.nva.useraccessservice.model.UserDto;
+import nva.commons.core.attempt.Try;
+import nva.commons.core.paths.UriWrapper;
+
+public class UserEntriesCreatorForPerson {
```
axthosarouris, line 22: Code was just moved from the UserSelectionUponLoginHandler to here

PR #166, opened 2022-04-21
Fix customer selection endpoint

axthosarouris:


user-access-service/src/test/java/no/unit/nva/database/IdentityServiceTest.java
```diff
@@ -84,24 +84,6 @@ public void init() {
         usersTable = enhancedClient.table(IdentityService.USERS_AND_ROLES_TABLE, UserDao.TABLE_SCHEMA);
     }
 
-    @Test
```
axthosarouris, line 87: only reformating of the file and renaming the method `getUserByCristinIdAndCristinOrgId` to `getUserByPersonCristinIdAndCustomerCristinId`

user-access-testing/src/main/java/no/unit/nva/database/LocalIdentityService.java
```diff
@@ -33,10 +33,9 @@
 import software.amazon.awssdk.services.dynamodb.model.TableDescription;
 import software.amazon.awssdk.services.dynamodb.model.TableStatus;
 
-public abstract class LocalIdentityService implements WithEnvironment {
+public class LocalIdentityService implements WithEnvironment {
```
axthosarouris, line 36: not abstract anymore because we want to initialize it in the handler test. We need both a local database for the customer service and for the identity service. We cannot have a common because we have to declare the database key schema. In  the future these two tables should be merged to one.

PR #171, opened 2022-04-28
UserEntriesCreatorForPerson takes only 3 parameters as input

axthosarouris:


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -93,6 +99,20 @@ public CognitoUserPoolPreTokenGenerationEvent handleRequest(CognitoUserPoolPreTo
         return input;
     }
 
+    private static String extractNin(Map<String, String> userAttributes) {
```
axthosarouris, line 102: moved outside the AuthenticationInformation because they are related to the event.

brinxmat (approved):

PR #196, opened 2022-05-03
Np 9101 persist workflow choice

torarnet:


axthosarouris (changes requested):
There should be a test in CreateCustomerHandler that verifies that the values is actually saved.

axthosarouris (approved):

brinxmat (approved):

PR #201, opened 2022-05-04
Create user refactoring part 2

axthosarouris:


user-access-handlers/src/main/java/no/unit/nva/handlers/AddUserHandler.java
```diff
@@ -1,49 +0,0 @@
-package no.unit.nva.handlers;
```
axthosarouris, line 1: Renamed to CreateUserHandler

brinxmat (approved):

PR #207, opened 2022-05-06
Np 9076 backend can create users

axthosarouris:


user-access-handlers/src/main/java/no/unit/nva/handlers/CreateUserHandler.java
```diff
@@ -90,11 +90,11 @@ private void authorize(RequestInfo requestInfo) throws ForbiddenException {
     }
 
     private boolean userIsNotAuthorized(RequestInfo requestInfo) {
-        var loggerInfoString = attempt(() -> JsonConfig.writeValueAsString(requestInfo)).orElseThrow();
```
axthosarouris, line 93: removed debugging logging

brinxmat (approved):

PR #227, opened 2022-05-12
fix-template

axthosarouris:


template.yaml
```diff
@@ -719,20 +720,6 @@ Resources:
             Method: put
             RestApiId: !Ref NvaCustomerApi
 
-  InitCustomerServiceHandler:
```
axthosarouris, line 722: We have only one init hander for the whole identity service

brinxmat (approved):

PR #238, opened 2022-07-18
NP9312-Create-app-client-for-DLR, client created

truhacevkir:
Ser det bra ut?

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -97,6 +97,8 @@ Resources:
       CognitoIdentityProviders:
         - ClientId: !Ref ApplicationClient
           ProviderName: !Sub 'cognito-idp.${AWS::Region}.amazonaws.com/${UserPool}'
+        - ClientId: !Ref DLRApplicationClient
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -1277,6 +1279,80 @@ Resources:
...
+        - custom:orgFeideDomain
+        - custom:firstName
+        - custom:lastName
+        - custom:accessRights
+        - custom:customerId
+        - custom:allowedCustomers
+        - custom:roles
+        - custom:nvaUsername
+        - custom:cristinId
+        - custom:topOrgCristinId
```
axthosarouris, line 1326: do you need these 3?

template.yaml
```diff
@@ -1277,6 +1279,80 @@ Resources:
...
```
axthosarouris, line 1352: ditto

template.yaml
truhacevkir: probably not, have removed them.

template.yaml
truhacevkir: removing these also

PR #244, opened 2022-08-19
Np 9329 dlr auth customer handling

truhacevkir:


customer-commons/src/main/java/no/unit/nva/customer/model/ApplicationDomain.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import java.net.URI;
+import java.util.Arrays;
+import nva.commons.core.SingletonCollector;
+
+public enum ApplicationDomain {
+    
+    NVA(URI.create("https://nva.org"));
```
brinxmat, line 11: We should change this -> but I don't know to what the final domain will be…at least nva.unit.no for now

user-access-service/build.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
     implementation libs.slf4j
     implementation libs.bundles.aws.lambda
 
+    testImplementation group: 'io.github.ganadist.sqlite4java', name: 'libsqlite4java-osx-aarch64', version: '1.0.392'
```
brinxmat, line 27: Can be removed

brinxmat (changes requested):

user-access-service/build.gradle
truhacevkir: done

customer-commons/src/main/java/no/unit/nva/customer/model/ApplicationDomain.java
truhacevkir: done

PR #247, opened 2022-09-26
pretoken generation handler only logs

axthosarouris:


buildSrc/src/main/groovy/nva.identity.service.java-common-conventions.gradle
```diff
@@ -37,10 +37,6 @@ tasks.named('test') {
     finalizedBy jacocoTestReport
 }
 
-test {
```
axthosarouris, line 40: removed from conventions because they cause environment variable to be set in a global level and not have overview where and why we need them

cognito-pre-token-generation-openid/build.gradle
```diff
@@ -45,4 +45,8 @@ test {
     environment "API_DOMAIN", "example.com"
     environment "CUSTOMERS_TABLE_NAME", "CustomersTable"
     environment "USERS_AND_ROLES_TABLE", "UsersAndRolesTable"
+    environment "ID_NAMESPACE", "https://www.example.org/namespace/"
```
axthosarouris, line 48: as a result of removing them from conventions file

cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -51,159 +51,143 @@
...
         final var usersForPerson = userCreator.createUsers(authenticationInfo);
-
+    
         final var roles = rolesPerCustomer(usersForPerson);
         authenticationInfo.updateCurrentCustomer();
         authenticationInfo.updateCurrentUser(usersForPerson);
-
-        final var accessRights = createAccessRights(usersForPerson);
+    
+        final var accessRights = createAccessRightsPerCustomer(usersForPerson);
```
axthosarouris, line 100: simplify method after removing legacy code

cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -51,159 +51,143 @@
...
-            .map(idNumber -> userCreator.collectPersonInformation(idNumber, personFeideIdentifier, orgFeideDomain))
-            .map(AuthenticationInformation::new)
-            .orElseThrow();
+                   .map(idNumber -> userCreator.collectPersonInformation(idNumber, personFeideIdentifier,
+                       orgFeideDomain))
+                   .map(AuthenticationInformation::new)
+                   .orElseThrow();
     }
-
-    private List<String> createAccessRights(List<UserDto> usersForPerson) {
```
axthosarouris, line 134: remove method after removing the legacy functionality of adding `USER@<customerId>` in the cognito groups for the e2e testing. Now we are using a dedicated endpoint for e2e tests

cognito-pre-token-generation-openid/src/test/java/no/unit/nva/cognito/AuthenticationScenarios.java
```diff
@@ -0,0 +1,119 @@
+package no.unit.nva.cognito;
```
axthosarouris, line 1: This class has been written with pair programming

template.yaml
```diff
@@ -1440,13 +1440,13 @@ Resources:
...
       Handler: no.unit.nva.cognito.UserSelectionUponLoginHandler::handleRequest
       Role: !GetAtt CreateUsersRole.Arn
-      AutoPublishAlias: live
-      Tracing: Active
-      DeploymentPreference:
-        Enabled: true
-        Type: AllAtOnce
-      ProvisionedConcurrencyConfig:
-        ProvisionedConcurrentExecutions: 1
+      #      Tracing: Active
```
axthosarouris, line 1443: disabling hot instances for now for easier debugging

PR #250, opened 2022-11-02
only include feide customer in allowedCustomers claim in cognito on f…

ketilaa:
…eide login

### DIFF
--- a/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
+++ b/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
@@ -27,6 +27,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import no.unit.nva.auth.AuthorizedBackendClient;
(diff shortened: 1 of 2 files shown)


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -219,9 +221,11 @@ private List<AttributeType> customerSelectionClaims(AuthenticationInformation au
         return List.of(currentCustomerClaim, currentTopLevelOrgClaim, usernameClaim, personAffiliationClaim);
     }
     
-    private String createAllowedCustomersString(Collection<CustomerDto> allowedCustomers) {
+    private String createAllowedCustomersString(Collection<CustomerDto> allowedCustomers, String feideDomain) {
         var result = allowedCustomers
                          .stream()
+                         .filter(customer -> !Objects.nonNull(feideDomain) || customer.getFeideOrganizationDomain()
```
brinxmat, line 227: For clarity's sake, extract the condition to a method — from what I understand, something like:
(note "Objects::isNull")

```
    private static boolean isNotFeideRequestOrIsFeideRequestWithCorrectDomain(String feideDomain, CustomerDto customer) {
        return isNull(feideDomain) || customer.getFeideOrganizationDomain().equals(feideDomain);
    }
```

ketilaa: Good point. Fixed.

ketilaa:
I have tested it in sandbox. I noticed the token endpoint return internal error some times. As that happens on dev too, I am pretty sure it is not related to these changes. Nothing to be found in pre-token-generation-handler logs either.

I have tested in sandbox with:

- feide login with two affiliations in Cristin, only one of them a customer in NVA
- non-feide login with two affiliations in Cristin, only one of them a customer in NVA.

I will try adding a new customer in NVA sandbox to match my second affiliation too.

ketilaa:
Seems to work as intended on sandbox.

brinxmat (approved):

PR #252, opened 2022-11-04
Remove customer selection claims if customer is ambigous during login 

ketilaa:
I selected the strategy that was already in use for the allowed customers claim to indicate that the claim has no value, which is to use the string value "null". I am open to other strategies like empty string or the string "empty", but I think "null" is familiar to all of us.

Customer selection claims covers the following claims:

- `custom:customerId`
- `custom:nvaUsername`
- `custom:topOrgCristinId`
- `custom:cristinId`

This change will basically force users with more than one active affiliation (employment) in Cristin to select the customer they want to log in to instead of assuming the same the user selected on previously logins when they log in not using Feide.

### DIFF
--- a/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
+++ b/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
@@ -43,6 +43,8 @@ import no.unit.nva.useraccessservice.usercreation.cristin.person.CristinClient;
 import nva.commons.core.Environment;
 import nva.commons.core.JacocoGenerated;
 import nva.commons.core.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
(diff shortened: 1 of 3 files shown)


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -211,19 +214,46 @@ private void addCustomerSelectionClaimsWhenUserHasOnePossibleLoginOrLoggedInWith
...
+    }
+
+    private Consumer<String> generateCustomerSelectionClaimsFromAuthentication(
+        AuthenticationInformation authenticationInfo,
+        List<AttributeType> claims) {
+        
+        return customerId -> claims.addAll(customerSelectionClaims(authenticationInfo, customerId));
+    }
+
+    private List<AttributeType> emptyCustomerSelectionClaims() {
```
brinxmat, line 232: overwriteAllCustomerSelectionClaimsWithNull

brinxmat (changes requested):

cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
ketilaa: Fixed and merged!

PR #261, opened 2022-11-25
fixed NullPointerException for Crisin persons without affiliations

ketilaa:
Cristin do not serialize empty arrays, so I made sure to handle that in deserialization.

### DIFF
--- a/user-creation-testing/src/main/java/no/unit/nva/useraccessservice/userceation/testing/cristin/MockPersonRegistry.java
+++ b/user-creation-testing/src/main/java/no/unit/nva/useraccessservice/userceation/testing/cristin/MockPersonRegistry.java
@@ -72,7 +72,7 @@ public class MockPersonRegistry {
     public String personWithoutAffiliations() {
         var nin = randomString();
 
-        createPersonWithAffiliations(nin, null);
+        createPersonWithoutAffiliations(nin);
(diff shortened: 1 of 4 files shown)


user-creation-testing/src/main/java/no/unit/nva/useraccessservice/userceation/testing/cristin/MockPersonRegistry.java
```diff
@@ -69,6 +69,14 @@ public CristinPerson getPerson(String nin) {
         return people.get(nin);
     }
 
+    public String personWithoutAffiliations() {
```
torarnet, line 72: personWithoutAffiliations method calls method createPersonWithAffiliations?

torarnet: or does that null value in parameter mean without?

ketilaa: Yes. I agree it is confusing naming. I will fix it.

torarnet (dismissed):

brinxmat (approved):

PR #264, opened 2023-01-02
added nullcheck

StigNorland:


StigNorland:
https://unit.atlassian.net/browse/NP-27646

customer/src/main/java/no/unit/nva/customer/get/GetCustomerDoiHandler.java
```diff
@@ -58,33 +61,41 @@ protected DoiAgentDto processInput(Void input, RequestInfo requestInfo, Context
...
+            var secret = getSecretsManagerDoiAgent().get(identifier);
+            if (nonNull(secret)) {
+                // if you want to show the password, add it here...
+                return new DoiAgentDto(secret)
+                           .addId(doiAgentId);
+            } else {
+                return new DoiAgentDto().addId(doiAgentId);
+            }
+        } catch (JsonProcessingException e) {
+            throw new BadRequestException(e.getMessage());
```
ketilaa, line 79: Is it a Bad Request (400)? The client did something wrong in the request to us? I would rather categorise this as an Internal Server Error. I would wrap the block in `Try.attempt().orElse(/* log and throw *(});`

ketilaa (changes requested):

customer/src/main/java/no/unit/nva/customer/get/GetCustomerDoiHandler.java
StigNorland: ok

ketilaa (approved):

PR #267, opened 2023-02-07
External users

sondrev:


docs/users-and-roles-swagger.yaml
```diff
@@ -164,6 +150,28 @@ paths:
         responses: {}
         httpMethod: POST
         type: aws_proxy
+  /external-users:
```
ketilaa, line 153: externalClients?

template.yaml
```diff
@@ -247,6 +251,17 @@ Resources:
                   - secretsmanager:GetRandomPassword
                   - secretsmanager:ListSecrets
                 Resource: '*'
+        - PolicyName: ExternalUserPoolClients
```
ketilaa, line 254: We encourage description for components we create in template.yaml

template.yaml
```diff
@@ -494,6 +509,26 @@ Resources:
         - DynamoDBCrudPolicy:
             TableName: !Ref NvaCustomersTable
 
+  CreateExternalUserClientHandler:
```
ketilaa, line 512: User -> Client?

user-access-handlers/src/test/java/no/unit/nva/handlers/CreateExternalUserClientHandlerTest.java
```diff
@@ -0,0 +1,108 @@
...
+    @BeforeEach
+    public void setup()  {
+        cognitoClient = Mockito.mock(CognitoIdentityProviderClient.class);
+        var externalUserService = new ExternalUserService(cognitoClient);
+
+        var response = CreateUserPoolClientResponse.builder().userPoolClient(
+            UserPoolClientType.builder().clientId(CLIENT_ID).clientSecret(CLIENT_SECRET).build()
+        ).build();
+
+        Mockito.when(cognitoClient.createUserPoolClient(any(CreateUserPoolClientRequest.class)))
```
ketilaa, line 53: static import of Mockito.mock makes it a lot easier to read

user-access-public-model/src/main/java/no/unit/nva/useraccessservice/model/CreateExternalUserRequest.java
```diff
@@ -0,0 +1,29 @@
...
+import static nva.commons.core.attempt.Try.attempt;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import no.unit.nva.identityservice.json.JsonConfig;
+
+public class CreateExternalUserRequest {
+
+    private static final String CLIENT_NAME = "clientName";
+
+    @JsonProperty(CLIENT_NAME)
+    private String clientName;
```
ketilaa, line 12: Would it be a good idea to allow for setting valid redirect urls? How does it work now? Does it default to *?

ketilaa (commented):
Did a quick read-through. Generally, I find ExternalClient* better than ExternalUser*. Bit I see that when creating auth client, createExternalClientClient is a bit weird...

I noticed some tests were not done through the handler, probably to get coverage. Have you explored your options on reaching all code through the handler invocation?

template.yaml
```diff
@@ -552,12 +587,17 @@ Resources:
         Format: '{ "apiId": "$context.apiId", "requestId": "$context.requestId", "requestTime": "$context.requestTime", "requestTimeEpoch": "$context.requestTimeEpoch", "httpMethod": "$context.httpMethod", "path": "$context.path", "status": "$context.status",  "error.message": "$context.error.message" }'
       StageName: Prod
       Auth:
+        AddDefaultAuthorizerToCorsPreflight: false
+        DefaultAuthorizer: CognitoAuthorizer
         Authorizers:
           CognitoAuthorizer:
+            AuthorizationScopes:
```
ketilaa, line 594: Are we still able to have different authorisers for different lambdas?

user-access-public-model/src/main/java/no/unit/nva/useraccessservice/model/CreateExternalUserRequest.java
sondrev: Redirect URLs I belive are for user authentication, and not needed for machine to machine.

template.yaml
sondrev: Do you mean different scopes? The old solution had 1 Authoriser defined in customer-swagger which was not actually used. 
Can we do a huddle on this?

template.yaml
sondrev: Does not seem like its possible to add a Description to the Policy object. I did however add a Description to the Lambda after your feedback.

template.yaml
```diff
@@ -19,6 +19,10 @@ Parameters:
     Type: String
     Default: nva-develop
     Description: The prefix for the Cognito domain that hosts the sign-up and sign-in pages for your application.
+  ExternalAuthenticationDomainPrefix:
+    Type: String
+    Default: nva-develop-ext
```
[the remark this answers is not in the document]
sondrev: This is a part of the URL that hosts the auth:
https://nva-develop-ext.auth.eu-west-1.amazoncognito.com/oauth2/token
It needs to unique, not only between our 4 enviorments, but across all of AWS. The idea is to do the same as the parameter above this one (AuthenticationDomainPrefix): We pass it down from nva-infrastructure template

PR #272, opened 2023-03-10
External clients in database

sondrev:


user-access-handlers/src/main/java/no/unit/nva/handlers/CreateExternalClientHandler.java
```diff
@@ -3,35 +3,83 @@
...
+    private void validateRequest(CreateExternalClientRequest input) throws BadRequestException {
+        if (input.getScopes() == null) {
+            throw new BadRequestException(MISSING_SCOPES);
+        }
+        if (input.getCustomer() == null) {
+            throw new BadRequestException(MISSING_CUSTOMER);
+        }
+        if (input.getClientName() == null) {
+            throw new BadRequestException(MISSING_CLIENT_NAME);
+        }
```
brinxmat, line 70: I wonder if we should collect the errors and return a complete error report?

brinxmat (dismissed):

PR #284, opened 2023-05-10
NP-44689-NP-44687-added-support-for-new-roles

anetteOlli:


user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceInitHandler.java
```diff
@@ -98,26 +101,39 @@ private List<RoleDto> createDefaultRoles() {
...
-                EDIT_OWN_INSTITUTION_PROJECTS,
-                EDIT_OWN_INSTITUTION_USERS))
-            .build();
+                                   .withAccessRights(List.of(EDIT_OWN_INSTITUTION_RESOURCES,
+                                                             EDIT_OWN_INSTITUTION_PROJECTS,
+                                                             EDIT_OWN_INSTITUTION_USERS))
+                                   .build();
+        var importCandidateCurator =
+            RoleDto.newBuilder()
+                .withRoleName("Import-candidate-curator")
```
StigNorland: Should use same naming convention on all Roles
either `Curator-import-candidate` or change `Curator-thesis` and `Curator-thesis-embargo`

StigNorland (changes requested):

user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceInitHandler.java
anetteOlli: yes, noticed just after I request your review myself.

StigNorland (approved):

PR #291, opened 2023-06-13
Np 44861 access rights

ketilaa:
The problem was that when user was updated during login, the updated user was not used in the login logic. Now fixed.

### DIFF
--- a/user-access-service/src/main/java/no/unit/nva/database/IdentityService.java
+++ b/user-access-service/src/main/java/no/unit/nva/database/IdentityService.java
@@ -34,8 +34,10 @@ public interface IdentityService {
     UserDto addUser(UserDto user) throws ConflictException;
     
     void addRole(RoleDto roleDto) throws ConflictException, InvalidInputException;
-    
-    void updateUser(UserDto user) throws NotFoundException;
+
+    void updateRole(RoleDto roleDto) throws InvalidInputException, NotFoundException;
+
+    UserDto updateUser(UserDto user) throws NotFoundException;
     
     RoleDto getRole(RoleDto input) throws NotFoundException;
     
--- a/user-access-service/src/test/java/no/unit/nva/database/IdentityServiceImplTest.java
+++ b/user-access-service/src/test/java/no/unit/nva/database/IdentityServiceImplTest.java
@@ -4,16 +4,21 @@ import static no.unit.nva.database.RoleService.ROLE_NOT_FOUND_MESSAGE;
 import static nva.commons.core.attempt.Try.attempt;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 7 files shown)


user-access-service/src/main/java/no/unit/nva/database/IdentityService.java
```diff
@@ -34,8 +34,10 @@ static IdentityService defaultIdentityService(DynamoDbClient dynamoDbClient) {
     UserDto addUser(UserDto user) throws ConflictException;
     
     void addRole(RoleDto roleDto) throws ConflictException, InvalidInputException;
-    
-    void updateUser(UserDto user) throws NotFoundException;
+
+    void updateRole(RoleDto roleDto) throws InvalidInputException, NotFoundException;
```
ketilaa, line 38: This is added just to make it possible from test. Currently, we have no API to manage roles. They are handled as system data.

user-creation/src/main/java/no/unit/nva/useraccessservice/usercreation/UserEntriesCreatorForPerson.java
```diff
@@ -104,8 +104,7 @@ private UserDto updateUserAffiliation(UserDto existingUser,
                                           Person person) throws NotFoundException {
         var affiliation = person.getConsistentUnitAffiliation(existingUser.getInstitutionCristinId());
         var updatedUser = existingUser.copy().withAffiliation(affiliation).build();
-        identityService.updateUser(updatedUser);
-        return updatedUser;
+        return identityService.updateUser(updatedUser);
```
ketilaa, line 107: Nor returning the updated user so that we detect new access rights during login.

truhacevkir (dismissed):

user-access-service/src/main/java/no/unit/nva/database/IdentityService.java
StigNorland: I was about to ask why you deleted updateUser, but lo and behold, below it was.

user-access-service/src/test/java/no/unit/nva/database/IdentityServiceImplTest.java
```diff
@@ -51,6 +56,24 @@ public void getRoleLogsWarningWhenNotFoundExceptionIsThrown() throws InvalidEntr
                    StringContains.containsString(ROLE_NOT_FOUND_MESSAGE));
     }
 
+    @Test
+    void shouldSucceedUpdatingAnExistingRole() throws InvalidInputException, ConflictException {
+        var serviceUnderTest = new IdentityServiceImpl(initializeTestDatabase());
```
StigNorland, line 61: the name `serviceUnderTest `doens't give me any clue as to why `databaseService  `cannot be used
consider renaming serviceUnderTest to something that hints as to why you need another instance of IdentiyService

StigNorland (changes requested):

user-access-service/src/test/java/no/unit/nva/database/IdentityServiceImplTest.java
ketilaa: Actually, it could be `databaseService`. Fixed.

truhacevkir (approved):

PR #297, opened 2023-06-20
NP-44893: access rights for managing projects

ketilaa:


user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceInitHandler.java
```diff
@@ -45,33 +32,36 @@ public class IdentityServiceInitHandler extends ApiGatewayHandler<Void, RoleList
...
+        this.roleSource = roleSource;
     }
     
     @Override
     protected RoleList processInput(Void input, RequestInfo requestInfo, Context context) {
-        var defaultRoles = createDefaultRoles()
+        var defaultRoles = roleSource.roles()
             .stream()
-            .map(attempt(this::addRole))
+            .map(attempt(this::addOrUpdateRole))
```
torarnet, line 54: Had an issue just now myself that error handling were not working correctly when using lambda annotation instead of regular method call in an attempt call. Not sure why, but have you tested what happens if this produce an exception?

user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceInitHandler.java
```diff
@@ -92,50 +82,14 @@ private void logError(Exception exception) {
...
+            identityService.addRole(role);
+            logger.info("Added role: {}", role);
+        } catch (ConflictException e) {
+            identityService.updateRole(role);
+            logger.info("Updated role: {}", role);
+        }
         return role;
     }
-    
-    private List<RoleDto> createDefaultRoles() {
```
torarnet, line 101: Better to have this in own class as you have done. My opinion is to keep as minimal code as possible in the handler itself, and rather delegate to other classes per purpose

torarnet (dismissed):
Looks good to me, but does RoleSource need to be reinitialized each time the handler is called?

user-access-handlers/src/main/java/no/unit/nva/handlers/data/DefaultRoleSource.java
```diff
@@ -0,0 +1,90 @@
...
+
+    protected static final String CURATOR_ROLE_NAME = "Curator";
+    protected static final String INSTITUTION_ADMIN_ROLE_NAME = "Institution-admin";
+    protected static final String CURATOR_IMPORT_CANDIDATE_ROLE_NAME = "Curator-Import-candidate";
+    protected static final String CURATOR_THESIS_ROLE_NAME = "Curator-thesis";
+    protected static final String CURATOR_THESIS_EMBARGO_ROLE_NAME = "Curator-thesis-embargo";
+    protected static final String APP_ADMIN_ROLE_NAME = "App-admin";
+    protected static final String EDITOR_ROLE_NAME = "Editor";
+
+    private static final RoleDto CREATOR_ROLE = RoleDto.newBuilder()
```
torarnet, line 31: Is creator the best name for a person with active employment?

user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceInitHandler.java
ketilaa: Agreed

user-access-handlers/src/main/java/no/unit/nva/handlers/data/DefaultRoleSource.java
ketilaa: Not really, but that is what it ended up as back in the days. Someone uses the term "Registrator" too. Anyway, changing it now might not be a good idea, but an overhaul of the roles, access rights and users sometime would definitely be a good idea. I am by no means assured that we have all the access rights that we should have, and that we actually do the correct checks in our handlers. PO ofter states "Curators should be allowed to ...", and often we just pick one of the existing access right given to curators only to check. Not optimal at all.

torarnet (approved):

PR #302, opened 2023-08-22
adding nvi role to app admin

truhacevkir:


template.yaml
```diff
@@ -1040,7 +1040,7 @@ Resources:
   CognitoAuthenticationDomainParameter:
     Type: AWS::SSM::Parameter
     Properties:
-      Name: !Sub '${ParameterNamePrefix}CognitoAuthenticationDomain'
+      Name: !Sub '${ParameterNamePrefix}CognitoAuthenIticationDomain'
```
ketilaa, line 1043: This change is probably not intentional?

ketilaa (changes requested):
I think you have an unintentional change in the template?

StigNorland (approved):

anetteOlli (approved):

PR #303, opened 2023-08-23
NP-32829-show-doiprefix-in-get-all-customers-handler

anetteOlli:


docs/customer-swagger.yaml
```diff
@@ -828,28 +841,9 @@ components:
...
-            archiveName: 'unirein'
-            cname: 'arkiv.unirein.no'
-            institutionDns: 'unirein.no'
-            feideOrganizationId: '111111111'
-            cristinId: 'https://api.cristin.no/v2/institutions/11110'
-            vocabularies:
-              - id: 'https://example.org/example#'
-                type: 'Vocabulary'
-                name: 'Example VOC'
-                active: 'default'
```
torarnet, line 847: Is all this some old swagger specs?

torarnet (approved):

docs/customer-swagger.yaml
anetteOlli: I don't think this swagger documentation matched the output from that endpoint even before I started working this branch.

PR #306, opened 2023-08-30
NP-45132-nvi-curator

hackon:


truhacevkir (approved):
Remember to run IdentityService init handler

PR #307, opened 2023-10-06
NP 45375 Fix rboInstitution not mapping to schema

joachimjorgensen:
Fixing rboInstution bug where the value was not put to the DynamoDB table.

Added new test to validate that rboInstitute value is persisted.

Fixed by recreating 
- getters
- setters
- hashcode
- equals

### DIFF
--- a/customer-commons/src/test/java/no/unit/nva/customer/service/impl/DynamoDBCustomerServiceTest.java
+++ b/customer-commons/src/test/java/no/unit/nva/customer/service/impl/DynamoDBCustomerServiceTest.java
@@ -1,9 +1,9 @@
 package no.unit.nva.customer.service.impl;
 
 import static no.unit.nva.customer.model.VocabularyStatus.ALLOWED;
-import static no.unit.nva.customer.testing.CustomerDataGenerator.randomDoiAgent;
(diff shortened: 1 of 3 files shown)


customer-commons/src/test/java/no/unit/nva/customer/service/impl/DynamoDBCustomerServiceTest.java
```diff
@@ -86,6 +88,19 @@ void updateExistingCustomerWithNewName() throws NotFoundException, InputExceptio
...
+    void shouldUpdateRboInstitutionWhenRboInstitutionIsSetToTrue()
+        throws NotFoundException, InputException, ConflictException {
+        var customer = newCustomerDto();
+        customer.setRboInstitution(false);
+        var createdCustomer = service.createCustomer(customer);
+        assertFalse(createdCustomer.isRboInstitution());
+
+        createdCustomer.setRboInstitution(true);
+        var updatedCustomer = service.updateCustomer(createdCustomer.getIdentifier(), createdCustomer);
+        assertTrue(updatedCustomer.isRboInstitution());
```
truhacevkir, line 101: Single assertion from false -> true is enough, but you can keep it as you want.

truhacevkir (approved):

monaullah (approved):

PR #308, opened 2023-11-08
Fixup rrs

brinxmat:
- Migrates erroneous naming of RightsRetentionStrategy in code
  - Migrates retentionStrategy to type
- Adds DB migration of the above


ketilaa (dismissed):
Smooth sailing ;)

PR #309, opened 2023-11-09
updated missing fields in swagger documentation for the Customer resource

ketilaa:


docs/customer-swagger.yaml
```diff
@@ -682,6 +682,32 @@ components:
...
+      type: object
+      properties:
+        type:
+          type: string
+          description: Always has the value 'RightsRetentionStrategy'
+        id:
+          type: string
+          format: uri
+          description: a link to the rights retention strategy for the customer
+    OverridableRightsRetentionStrategy:
```
StigNorland, line 701: What happens if the uri is outside our domain?

docs/customer-swagger.yaml
```diff
@@ -682,6 +682,32 @@ components:
...
+      type: object
+      properties:
+        type:
+          type: string
+          description: Always has the value 'OverridableRightsRetentionStrategy'
+        id:
```
StigNorland, line 707: same same

torarnet (approved):

PR #315, opened 2023-12-12
Adding publication types (categories) to customer api.

EyobSikt:
Adding array of publication types (categories) to customer api.


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -66,6 +67,7 @@ public class CustomerDao implements Typed {
     private Sector sector;
     @JsonAlias("rightRetentionStrategy")
     private RightsRetentionStrategyDao rightsRetentionStrategy;
+    private List<PublicationTypes> publicationTypes;
```
ketilaa, line 70: I think the naming is a bit off here. It becomes unclear what it is. `allowFileUploadForPublicationInstances`, or simply `allowFileUploadFor` would probably be better.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -231,6 +233,14 @@ public void setSector(Sector sector) {
         this.sector = sector;
     }
 
+    public List<PublicationTypes> getPublicationTypes() {
```
brinxmat, line 236: A set, as above?

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -231,6 +233,14 @@ public void setSector(Sector sector) {
...
+        return publicationTypes;
+    }
+
+    public void setPublicationTypes(List<PublicationTypes> publicationTypes) {
```
brinxmat, line 240: For user-friendliness, make the argument a `Collection<PublicationTypes>`, assigning via `new HashSet(publicationTypes)`

customer-commons/src/main/java/no/unit/nva/customer/model/PublicationTypes.java
```diff
@@ -0,0 +1,75 @@
+package no.unit.nva.customer.model;
+
+public enum PublicationTypes {
```
brinxmat, line 3: There's likely an issue here because we have at some point a comparison with the actual manifestation of the types (the types are actually URIs in the data, this isn't apparent because of the JSON-LD wrapping of the data). 

Add e.g. ACADEMIC_ARTICLE("AcademicArticle") and return the string value as the JSON serialization. There is an easy way to make this change, contact me if you need help :)

customer/src/test/java/no/unit/nva/customer/create/CreateCustomerHandlerTest.java
```diff
@@ -148,6 +149,20 @@ void shouldReturnConflictErrorWhenTryingToCreateCustomerForInstitutionThatIsAlea
...
+            CustomerDto.builder()
+                .withName("New Customer")
+                .withPublicationTypes(randomPublicationTypes)
+                .build();
+        var requestBody = CreateCustomerRequest.fromCustomerDto(customerDto);
+        var response = executeRequest(requestBody, CustomerDto.class);
+        var actualResponseBody = CustomerDto.fromJson(response.getBody());
+        assertThat(actualResponseBody.getPublicationTypes(), is(equalTo(randomPublicationTypes)));
+    }
+
```
brinxmat, line 165: Can you add a test to ensure that the publication response actually serializes to the expected format? Since we are applying data from Publication, these values need to be in the same namespace since we will at some point want to compare the values. It's useful for users to be able to compare real values (i.e. URIs).

docs/customer-swagger.yaml
```diff
@@ -779,6 +779,21 @@ components:
             - $ref: '#/components/schemas/NullRightsRetentionStrategy'
             - $ref: '#/components/schemas/RightsRetentionStrategy'
             - $ref: '#/components/schemas/OverridableRightsRetentionStrategy'
+        publicationTypes:
+          type: array
+          items:
+            type: string
+            enum:
+              - ACADEMIC_ARTICLE
```
brinxmat, line 787: The list isn't exhaustive and needs updating to reflect the change to string.

docs/customer-swagger.yaml
```diff
@@ -895,6 +910,16 @@ components:
         rightsRetentionStrategy:
           type: OverridableRightsRetentionStrategy
           id: 'https://unirein.no/rights-retention-strategy.html'
+        publicationTypes:
+          - ACADEMIC_ARTICLE
```
brinxmat, line 914: Ditto

brinxmat (changes requested):

ketilaa (changes requested):
Better naming for field `publicationTypes` and enum `PublicationTypes` is required.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -43,8 +43,11 @@ public class CustomerDao implements Typed {
     public static final String CRISTIN_ID = "cristinId";
     public static final String TYPE = "Customer";
     public static final Set<VocabularyDao> EMPTY_VALUE_ACCEPTABLE_BY_DYNAMO = null;
+    public static final Set<PublicationInstanceTypes>
+        PUBLICATION_INSTANCE_TYPES_EMPTY_VALUE_ACCEPTABLE_BY_DYNAMO = null;
     public static final TableSchema<CustomerDao> TABLE_SCHEMA = TableSchema.fromClass(CustomerDao.class);
     public static final String VOCABULARIES_FIELD = "vocabularies";
+    public static final String PUBLICATION_TYPES_FIELD = "allowFileUploadFor";
```
ketilaa, line 50: Name of constant should relate to field name. ALLOW_FILE_UPLOAD_FOR_FIELD is the appropriate name.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -66,9 +69,11 @@ public class CustomerDao implements Typed {
     private Sector sector;
     @JsonAlias("rightRetentionStrategy")
     private RightsRetentionStrategyDao rightsRetentionStrategy;
+    private Set<PublicationInstanceTypes> allowFileUploadFor;
 
     public CustomerDao() {
         vocabularies = EMPTY_VALUE_ACCEPTABLE_BY_DYNAMO;
+        allowFileUploadFor = PUBLICATION_INSTANCE_TYPES_EMPTY_VALUE_ACCEPTABLE_BY_DYNAMO;
```
ketilaa, line 76: Same, ALLOW_FILE_UPLOAD_FOR_EMPTY_VALUE_ACCEPTABLE_BY_DYNAMO is the appropriate name.

docs/customer-swagger.yaml
```diff
@@ -779,6 +779,83 @@ components:
             - $ref: '#/components/schemas/NullRightsRetentionStrategy'
             - $ref: '#/components/schemas/RightsRetentionStrategy'
             - $ref: '#/components/schemas/OverridableRightsRetentionStrategy'
+        allowFileUploadFor:
+          type: array
+          items:
+            type: string
+            enum:
+              - ACADEMIC_ARTICLE
+              - ACADEMIC_CHAPTER
```
ketilaa, line 788: These should now follow naming pattern in enum: AcademicChapter, AcademicLitteratureReview and so on.

ketilaa (changes requested):

ketilaa (dismissed):

ketilaa (approved):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -231,6 +234,14 @@ public void setSector(Sector sector) {
         this.sector = sector;
     }
 
+    public Set<PublicationInstanceTypes> getAllowFileUploadFor() {
+        return allowFileUploadFor;
+    }
+
+    public void setAllowFileUploadFor(Set<PublicationInstanceTypes> allowFileUploadFor) {
```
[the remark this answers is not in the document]
brinxmat: Again `new HashSet(types)`

customer/src/main/java/no/unit/nva/customer/create/CreateCustomerRequest.java
```diff
@@ -206,6 +211,14 @@ public void setSector(Sector sector) {
         this.sector = sector;
     }
 
+    public Set<PublicationInstanceTypes> getAllowFileUploadFor() {
```
brinxmat, line 214: Types

brinxmat (changes requested):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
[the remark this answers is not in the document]
ketilaa: Why do you want to allow code using CustomerDto to send in anything that might/will change when set internally (like removing duplicates)? Can you elaborate on how this makes the setter more user-friendly? Personally, I would also have used the Set interface in this case.

brinxmat (approved):

PR #324, opened 2024-01-03
curator migration

sondrev:


ketilaa (approved):
What about the old "curator" role itself? Will it be deleted from the database in existing systems?

PR #329, opened 2024-01-11
added support to search for multiple roles

sondrev:


user-access-handlers/src/main/java/no/unit/nva/handlers/ListByInstitutionHandler.java
```diff
@@ -40,13 +46,13 @@ protected Integer getSuccessStatusCode(Void body, UserList output) {
...
-        var role = input.getQueryParameterOpt("role");
-        var userName = input.getQueryParameterOpt("name");
+        var roles = new HashSet<>(input.getMultiValueQueryParameter(QUERY_PARAM_ROLE));
+        var userName = input.getQueryParameterOpt(QUERY_PARAM_NAME);
         var users =
             databaseService
                 .listUsers(institutionId).stream()
                 .filter(user -> likeUserOrEmpty(user, userName))
-                .filter(user -> hasRoleOrEmpty(user, role))
+                .filter(user -> roles.isEmpty() || hasOneRole(user, roles))
```
StigNorland, line 55: This is my old code, I understand what it does, but not why....

user-access-handlers/src/main/java/no/unit/nva/handlers/ListByInstitutionHandler.java
```diff
@@ -63,9 +69,8 @@ private boolean likeUserOrEmpty(UserDto userDto, Optional<String> userName) {
                    .orElse(true);
     }
 
-    private boolean hasRoleOrEmpty(UserDto userDto, Optional<String> roleName) {
-        return roleName
-                   .map(s -> userDto.getRoles().stream().anyMatch(f -> f.getRoleName().equals(s)))
-                   .orElse(true);
+    private boolean hasOneRole(UserDto userDto, Set<String> checkedRoles) {
```
StigNorland, line 72: This returns true if there are more than zero matches between roles and user roles, but is that enough to grant access?

PR #344, opened 2024-02-19
all curators should have resource-accessright

sondrev:


user-access-handlers/src/main/java/no/unit/nva/handlers/data/DefaultRoleSource.java
```diff
@@ -81,13 +72,15 @@ public class DefaultRoleSource implements RoleSource {
 
     private static final RoleDto CURATOR_THESIS_ROLE = RoleDto.newBuilder()
                                                            .withRoleName(CURATOR_THESIS_ROLE_NAME)
-                                                           .withAccessRights(List.of(MANAGE_DEGREE))
+                                                           .withAccessRights(List.of(MANAGE_RESOURCES_STANDARD,
```
sondrev, line 75: Since Nvi,Doi,support,publishing-curators have MANAGE_RESOURCES_STANDARD i assumed we also should have it for the last 2 curator-types @ketilaa

ketilaa: I am not sure that is the intention. I will check it with Jan Erik.

ketilaa (approved):

PR #346, opened 2024-02-26
feat: get external client without passing client id in path #NP-46258

torbjokv:


docs/users-and-roles-swagger.yaml
```diff
@@ -204,6 +204,29 @@ paths:
         responses: {}
         httpMethod: POST
         type: aws_proxy
+    get:
+      summary: Get a stored client and its associated customer
+      security:
+        - CognitoUserPool: [ "https://api.nva.unit.no/scopes/backend", "https://api.nva.unit.no/scopes/frontend", "aws.cognito.signin.user.admin" ]
```
sondrev, line 210: -ExternalUserPool: []

sondrev: ExternalCognitoUserPool:
      type: apiKey
      name: Authorization
      in: header
      x-amazon-apigateway-authtype: cognito_user_pools
      x-amazon-apigateway-authorizer:
        type: cognito_user_pools
        providerARNs:
          - Ref: ExternalCognitoAuthorizerArn

docs/users-and-roles-swagger.yaml
```diff
@@ -204,6 +204,36 @@ paths:
         responses: {}
         httpMethod: POST
         type: aws_proxy
+  /external-client-userinfo:
```
sondrev, line 207: /external-client/userinfo ?

sondrev (approved):

docs/users-and-roles-swagger.yaml
torbjokv: yeah, was not sure about this. what you suggest is in conflict with /external-clients/{clientId}  in my head, how about  /clientinfo ?

truhacevkir (approved):

ketilaa (approved):

PR #347, opened 2024-03-05
Np 46159 attaching provided subunits when updating user

truhacevkir:
Update CreateUser endpoint to user creation only.

- Adding `viewingScope` to `CreateUserRequest` to be able to create user with viewing scope (fixing bug)
- Removing roles update from `CreateUser` handler
- Updating openapi schema for `CreateUser` endpoint 
- Returning `200 OK` WITHOUT OVERWRITING user data when attempting to create existing user.


user-access-handlers/src/main/java/no/unit/nva/handlers/models/ImpersonationRequest.java
```diff
@@ -6,10 +6,10 @@
 
 public class ImpersonationRequest {
 
-    protected static final String NATIONAL_IDENTITY_NUMBER_FIELD = "nin";
+    protected static final String NATIONAL_IDENTITY_NUMBER_FIELD = "nationalIdentityNumber";
 
     @JsonProperty(NATIONAL_IDENTITY_NUMBER_FIELD)
-    @JsonAlias("nin")
+    @JsonAlias("nationalIdentityNumber")
```
truhacevkir, line 12: Removing alias, frontend uses `nationalIdentityNumber`

user-access-handlers/src/test/java/no/unit/nva/handlers/models/CreateUserRequestTest.java
```diff
@@ -1,31 +0,0 @@
...
-import static no.unit.nva.testutils.RandomDataGenerator.randomString;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.core.Is.is;
-import static org.hamcrest.core.IsEqual.equalTo;
-import static org.hamcrest.core.IsInstanceOf.instanceOf;
-import java.io.IOException;
-import no.unit.nva.identityservice.json.JsonConfig;
-import org.junit.jupiter.api.Test;
-
-class CreateUserRequestTest {
```
truhacevkir, line 12: No needed longer, this class has been testing deserialization of json alias - alias is removed now

user-access-handlers/src/main/java/no/unit/nva/handlers/models/ImpersonationRequest.java
sondrev, line 12: You can just delete the alias as its the same as the property. 

Also the frontend have not implemented Impersonation so no need to change this?

user-creation/src/main/java/no/unit/nva/useraccessservice/usercreation/person/NationalIdentityNumber.java
```diff
@@ -8,6 +8,10 @@ public NationalIdentityNumber(final String nin) {
         this.nin = nin;
     }
 
+    public static NationalIdentityNumber fromString(final String nin) {
+        return new NationalIdentityNumber(nin);
+    }
```
sondrev, line 13: How is this different from the construtor? Both are public. Maybe make the constructor private, or use the constructor directly instead?

docs/users-and-roles-swagger.yaml
```diff
@@ -360,6 +381,25 @@ components:
             $ref: '#/components/schemas/AccessRight'
         viewingScope:
           $ref: '#/components/schemas/ViewingScope'
+    CreateUserRequest:
+      properties:
+        type:
+          type: string
+          required: true
+          enum:
+            - 'CreateUserRequest'
```
sondrev, line 390: You dont use this "type" in CreateUserREquest. Is it needed?

truhacevkir: It seems like we provide object types in given swagger schema for almost all objects, but value for those is set to string, which is in my opinion strange/not correct. So when frontenders will consume post vs put endpoints, they can differentiate between request objects, in this case CreateUserRequest vs User:


```
{
  "type": "CreateUserRequest",
  "nin": "string",
  "customerId": "string",
  "roles": [
    {
      "type": "string",
      "rolename": "string",
      "accessRight": [
        "APPROVE_DOI_REQUEST"
      ]
    }
  ],
  "viewingScope": {
    "includedUnits": [
      "string"
    ]
  }
} 
```

vs

```
{
  "type": "User",
  "username": "string",
  "givenName": "string",
  "familyName": "string",
  "institution": "string",
  "roles": [
    {
      "type": "string",
      "rolename": "string",
      "accessRight": [
        "APPROVE_DOI_REQUEST"
      ]
    }
  ],
  "accessRight": [
    "APPROVE_DOI_REQUEST"
  ],
  "viewingScope": {
    "includedUnits": [
      "string"
    ]
  }
}
```

ketilaa (approved):

PR #349, opened 2024-03-06
removing array validation from viewingScope

truhacevkir:
- Removing validation of empty array from viewingScope (Array entries are still validated)
- Removed "type" variable from ViewingScope - using Jackson annotation instead

### DIFF
--- a/user-access-public-model/src/test/java/no/unit/nva/useraccessservice/model/ViewingScopeTest.java
+++ b/user-access-public-model/src/test/java/no/unit/nva/useraccessservice/model/ViewingScopeTest.java
@@ -2,13 +2,15 @@ package no.unit.nva.useraccessservice.model;
 
 import static no.unit.nva.RandomUserDataGenerator.randomCristinOrgId;
 import static no.unit.nva.testutils.RandomDataGenerator.randomString;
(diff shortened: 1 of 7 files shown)


user-access-public-model/src/test/java/no/unit/nva/useraccessservice/model/ViewingScopeTest.java
```diff
@@ -48,14 +50,26 @@ void shouldSerializeAndDeserialize() throws BadRequestException {
...
+    void shouldDeserializeNullViewingScopeToViewingScopeWithEmptyList() throws JsonProcessingException {
+        var json = """
+            {
+              "type": "ViewingScope"
+            }
+            """;
+        var viewingScope = JsonUtils.dtoObjectMapper.readValue(json, ViewingScope.class);
+        assertThat(viewingScope.getExcludedUnits(), is(emptyIterable()));
+        assertThat(viewingScope.getIncludedUnits(), is(emptyIterable()));
+    }
```
truhacevkir, line 71: This is the only relevant part.

ketilaa: What if `user.viewingScope` is `null`?

ketilaa (approved):

PR #353, opened 2024-04-10
migrating serviceCenterUri to ServiceCenter

truhacevkir:
Migrating serviceCenterUri to:

```
"serviceCenter": {
                "uri": "https://www.example.com/vplpKxc5Iz8I",
                "name": "ragknjfsngkj"
              }
```

Is needed for GUI to able to show custom text-link instead of plain url


ketilaa (dismissed):

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -743,4 +723,70 @@ public String toString() {
             return toJsonString();
         }
     }
+
+    @DynamoDbBean
+    public static class ServiceCenterDao implements JsonSerializable {
+
+        private URI uri;
+        private String text;
```
brinxmat, line 731: What is "text" here? The display name? Is it possible that this is multilingual?

brinxmat: Conventionally, name is the "name", i.e. it is only called this thing…while label can be multilingual

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDto.java
```diff
@@ -203,12 +208,27 @@ public void setRorId(URI rorId) {
...
+    @JsonProperty("serviceCenter")
+    public void setServiceCenter(Object serviceCenter) {
+        if (serviceCenter instanceof Map<?,?>) {
+            var map = (HashMap) serviceCenter;
+            var uri = attempt(() -> URI.create(map.get("uri").toString())).orElse(failure -> null);
+            var text = attempt(() -> map.get("text").toString()).orElse(failure -> null);
+            this.serviceCenter = new ServiceCenter(uri, text);
+        } else if (serviceCenter instanceof String || serviceCenter instanceof URI) {
+            this.serviceCenter = new ServiceCenter(URI.create(serviceCenter.toString()), null);
+        } else if (serviceCenter instanceof ServiceCenter) {
```
brinxmat, line 227: We maybe aren't using java21, so my comment about variables is rhetorical

PR #354, opened 2024-04-11
adding allowsGeneralSupprot variable to customer

truhacevkir:


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -126,13 +129,21 @@ public void setInactiveFrom(Instant inactiveFrom) {
         this.inactiveFrom = inactiveFrom;
     }
 
+    public boolean isAllowsGeneralSupport() {
```
truhacevkir, line 132: There may be better name for this variable so getter is more readable.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -67,6 +68,7 @@ public class CustomerDao implements Typed {
     private DoiAgentDao doiAgent;
     private boolean nviInstitution;
     private boolean rboInstitution;
+    private Boolean allowsGeneralSupport;
```
brinxmat, line 71: A boolean value, when we probably want something else, but OK.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
brinxmat: `allowsGeneralSupport` is probably what we are looking for — does this conflict? Maybe

brinxmat (commented):
In general, the types of support that are provided need to be analysed — I guess that GS, DOI and so on are checkable

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
ketilaa: or `generalSupportEnabled`, which makes the setter look like this: `isGeneralSupportEnabled()`

ketilaa (commented):
Added a few comments

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
truhacevkir: I agree, but pmd does not like that getter for boolean variable is named without `is` in method name.

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
truhacevkir: This is made to made to be able to set this field to `true` for all all current customers. Otherwise it will be set to `false` by default. I can "map" it change it boolean where everything is migrated

brinxmat (approved):

PR #355, opened 2024-04-11
Rewrite existing lambda to customer batch scan

truhacevkir:


customer-commons/src/main/java/no/unit/nva/customer/service/CustomerService.java
```diff
@@ -23,5 +23,7 @@ public interface CustomerService {
     CustomerDto updateCustomer(UUID identifier, CustomerDto customer) throws InputException, NotFoundException;
 
     CustomerDto getCustomerByCristinId(URI cristinId) throws NotFoundException;
+    CustomerDto refreshCustomer(CustomerDto customer) throws NotFoundException;
```
ketilaa: Probably not part of interface, but rather a private method in the implementation?

customer-commons/src/test/java/no/unit/nva/customer/service/impl/DynamoDBCustomerServiceTest.java
```diff
@@ -91,6 +92,18 @@ void updateExistingCustomerWithNewName() throws NotFoundException, InputExceptio
...
     }
 
+    @Test
+    void shouldRefreshCustomers() throws ConflictException, NotFoundException {
+        service.createCustomer(newActiveCustomerDto());
+        assertDoesNotThrow(() -> service.refreshCustomers());
+    }
+
+    @Test
+    void shouldRefreshCustomer() throws ConflictException, NotFoundException {
```
ketilaa: Probably not needed as it is something we never do.

ketilaa (approved):

PR #358, opened 2024-04-18
allowing manage-own-affiliation to update vocabularies

truhacevkir:
Implemented:

- Allowing user with access right manage-own-affiliation to update vocabularies
- Allowing user with access right manage-own-affiliation to create vocabularies
- Removed serviceCenter migration code
- Removed serviceCenterUri as it has been migrated to serviceCenter in all environments
- Removed migration code for generalSupportEnabled (made getter return actual value instead of always returning true)

### DIFF
--- a/customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
+++ b/customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
@@ -17,6 +17,7 @@ import java.util.UUID;
 import java.util.stream.Collectors;
 import no.unit.nva.commons.json.JsonSerializable;
 import no.unit.nva.customer.model.CustomerDto.DoiAgentDto;
+import no.unit.nva.customer.model.CustomerDto.ServiceCenter;
 import no.unit.nva.customer.model.dynamo.converters.DoiAgentConverter;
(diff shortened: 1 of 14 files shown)


customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -453,7 +442,7 @@ public String toString() {
     }
 
     public ServiceCenterDao getServiceCenter() {
-        return new ServiceCenterDao(serviceCenterUri, null);
+        return nonNull(serviceCenter) ? serviceCenter : new ServiceCenterDao();
```
truhacevkir, line 445: removing migration code

customer-commons/src/main/java/no/unit/nva/customer/model/CustomerDao.java
```diff
@@ -130,7 +129,7 @@ public void setInactiveFrom(Instant inactiveFrom) {
     }
 
     public boolean isGeneralSupportEnabled() {
-        return true;
+        return generalSupportEnabled;
     }
```
truhacevkir, line 133: Removing migration here as well

ketilaa (approved):

PR #365, opened 2024-06-27
Np 47131 prevent persistence of invalid role names

truhacevkir:
We had a case where "old" role has still been attached to user in NVA. It caused that user had access rights we did not know about, cause role and all belongings access rights have not been removed/ updated when we updated user roles via api. 

To fix this case and all future cases we decided to make RoleName to enum, so all unknown values will fail on serialization. 

- Since roleName has been string before, had to update a lot of tests. 
- RoleName is still stored as String in database and all enum values have the same role names as before.

### DIFF
--- a/user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceMigrateCuratorHandler.java
+++ b/user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceMigrateCuratorHandler.java
@@ -1,6 +1,7 @@
 package no.unit.nva.handlers;
 
 import static no.unit.nva.handlers.data.DefaultRoleSource.NVI_CURATOR_ROLE;
+import static no.unit.nva.useraccessservice.model.RoleName.DEPRECATED_NVI_CURATOR;
 import static nva.commons.core.attempt.Try.attempt;
 import com.amazonaws.services.lambda.runtime.Context;
 import java.net.HttpURLConnection;
@@ -18,7 +19,6 @@ import org.slf4j.LoggerFactory;
 public class IdentityServiceMigrateCuratorHandler extends ApiGatewayHandler<Void, Void> {
 
     private static final Logger logger = LoggerFactory.getLogger(IdentityServiceMigrateCuratorHandler.class);
-    public static final String LEGACY_NVI_CURATOR_NAME = "Nvi-curator";
 
     private final IdentityService identityService;
 
@@ -47,7 +47,7 @@ public class IdentityServiceMigrateCuratorHandler extends ApiGatewayHandler<Void
     }
     
     private boolean userIsLegacyNviCurator(UserDto user) {
-        return user.getRoles().stream().map(RoleDto::getRoleName).anyMatch(LEGACY_NVI_CURATOR_NAME::equals);
+        return user.getRoles().stream().map(RoleDto::getRoleName).anyMatch(DEPRECATED_NVI_CURATOR::equals);
     }
 
     private UserDto attemptUpdateRolesForUser(UserDto user) {
@@ -65,7 +65,7 @@ public class IdentityServiceMigrateCuratorHandler extends ApiGatewayHandler<Void
 
     private Set<RoleDto> updateRoleSet(Set<RoleDto> roleSet) {
         roleSet.add(NVI_CURATOR_ROLE);
-        roleSet.removeIf(role -> LEGACY_NVI_CURATOR_NAME.equals(role.getRoleName()));
+        roleSet.removeIf(role -> DEPRECATED_NVI_CURATOR.equals(role.getRoleName()));
         return roleSet;
     }
 }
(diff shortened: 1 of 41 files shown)


user-access-handlers/src/main/java/no/unit/nva/handlers/IdentityServiceMigrateCuratorHandler.java
```diff
@@ -65,7 +65,7 @@ private UserDto updateRolesForUser(UserDto user) throws NotFoundException {
 
     private Set<RoleDto> updateRoleSet(Set<RoleDto> roleSet) {
         roleSet.add(NVI_CURATOR_ROLE);
-        roleSet.removeIf(role -> LEGACY_NVI_CURATOR_NAME.equals(role.getRoleName()));
+        roleSet.removeIf(role -> DEPRECATED_NVI_CURATOR.equals(role.getRoleName()));
```
truhacevkir, line 68: Had to add "old" nvi-curator role to be able to run migration test

user-access-internal-model/src/test/java/no/unit/nva/useraccessservice/dao/RoleDbTest.java
```diff
@@ -112,38 +110,38 @@ void buildWithoutRoleNameShouldThrowException() {
...
-        assertThat(exception.getMessage(), containsString(EMPTY_ROLE_NAME_ERROR));
-    }
+//    @ParameterizedTest(name = "setPrimaryHashKey throws exception when input is:\"{0}\"")
+//    @NullAndEmptySource
+//    @ValueSource(strings = {" ", "\t", "\n", "\r"})
+//    void setPrimaryHashKeyThrowsExceptionWhenInputIsBlankOrNullString(String blankString) {
+//        Executable action = () -> RoleDb.newBuilder().withName(blankString).build();
+//        InvalidEntryInternalException exception = assertThrows(InvalidEntryInternalException.class, action);
+//        assertThat(exception.getMessage(), containsString(EMPTY_ROLE_NAME_ERROR));
+//    }
```
truhacevkir, line 132: Following test is not longer relevant

user-access-internal-model/src/test/java/no/unit/nva/useraccessservice/dao/UserDaoTest.java
```diff
@@ -174,42 +175,42 @@ void shouldConvertToDtoAndBackWithoutInformationLoss() {
...
+//        throws InvalidEntryInternalException {
+//        RoleDto invalidRole = RoleDto.newBuilder().withRoleName(SOME_ROLENAME).build();
+//        invalidRole.setRoleName(invalidRoleName);
+//        List<RoleDto> invalidRoles = Collections.singletonList(invalidRole);
+//        UserDto userWithInvalidRole = UserDto.newBuilder().withUsername(SOME_USERNAME).withRoles(invalidRoles).build();
+//
+//        Executable action = () -> UserDao.fromUserDto(userWithInvalidRole);
+//        RuntimeException exception = assertThrows(RuntimeException.class, action);
+//        assertThat(exception.getCause(), is(instanceOf(InvalidEntryInternalException.class)));
+//    }
```
truhacevkir, line 207: Unsure if we still need these tests, all the unknown enum values will from now on be illegal.

user-access-public-model/src/test/java/no/unit/nva/useraccessservice/model/RoleDtoTest.java
```diff
@@ -50,40 +48,33 @@ void roleDtoShouldHaveABuilder() {
...
     }
 
-    @ParameterizedTest(name = "builder should throw exception when rolename is:\"{0}\"")
-    @NullAndEmptySource
-    @ValueSource(strings = {"", " "})
-    void builderShouldNotAllowEmptyRoleName(String rolename) {
-        Executable action = () -> RoleDto.newBuilder().withRoleName(rolename).build();
-        assertThrows(InvalidEntryInternalException.class, action);
-    }
-
```
truhacevkir, line 74: Not relevant test longer

truhacevkir:
From now we will end up with:

java.lang.IllegalArgumentException: 
Unable to convert string value 'Curator' to enum type 'class no.unit.nva.useraccessservice.model.RoleName'

And return internal server error via api.

Before:
Not longer "existent" role has been served via api and user got granted access rights to given role:
```
{
    "rolename" : "Curator",
    "accessRights" : [ "MANAGE_PUBLISHING_REQUESTS", "MANAGE_OWN_AFFILIATION", "MANAGE_DOI", "MANAGE_RESOURCES_ALL", "MANAGE_RESOURCES_STANDARD" ],
    "type" : "Role"
  }
```

sondrev (approved):
Looks good

anetteOlli (commented):
Too long, didn't read.

Quick question what will happen to a user that has one of the legacy userRoles when they log in now? Do we need to wash our databases before we roll this to prod?

truhacevkir:
> Too long, didn't read. (I am sure it just a lot of refactor from a string to enum so its fine).
> 
> Quick question what will happen to a user that has one of the legacy userRoles when they log in now? Do we need to wash our databases before we roll this to prod?

All "old" roles will be removed from users.

anetteOlli (dismissed):

PR #380, opened 2024-10-24
new instance type AcademicCommentary added, and improved logging on …

ketilaa:
…failure to create user

### DIFF
--- a/user-access-handlers/src/main/java/no/unit/nva/handlers/CreateUserHandler.java
+++ b/user-access-handlers/src/main/java/no/unit/nva/handlers/CreateUserHandler.java
@@ -232,7 +232,7 @@ public class CreateUserHandler extends HandlerWithEventualConsistency<CreateUser
         if (exception instanceof ConflictException) {
             return (ConflictException) exception;
         } else {
-            LOGGER.error(failure.getException().getMessage());
+            LOGGER.error("Failed to create or update user!", failure.getException());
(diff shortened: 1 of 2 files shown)


sondrev (approved):

user-access-handlers/src/main/java/no/unit/nva/handlers/CreateUserHandler.java
```diff
@@ -232,7 +232,7 @@ public static ApiGatewayException castToCorrectRuntimeException(Failure<?> failu
         if (exception instanceof ConflictException) {
             return (ConflictException) exception;
         } else {
-            LOGGER.error(failure.getException().getMessage());
+            LOGGER.error("Failed to create or update user!", failure.getException());
```
torarnet, line 235: We want the whole stacktrace, not just the message?

torarnet: Or does it gets formatted by the logger?

ketilaa: We want the source of the problem hidden in the stacktrace, yes. Currently it logs this:

ERROR - Expected a single value, but 0 were found

torarnet (approved):
Seems codacy complains, but approved otherwise

PR #384, opened 2024-11-11
NP-47873/terms-conditions

StigNorland:


docs/users-and-roles-swagger.yaml
```diff
@@ -164,7 +164,32 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/UpsertTermsConditionsRequest'
+      responses:
+        '202':
+          description: Successful request for updating the user. User will be eventually updated
+          headers:
+            Location:
+              schema:
+                type: string
+              description: Relative path to the user URI
```
truhacevkir, line 186: I see that following swagger file misses a lot of respons codes for multiple endpoints. I think you should add more possible responses here, like 401? 404? 502?

user-access-handlers/src/main/java/no/unit/nva/handlers/UpdatePersonTermsConditionsHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    protected TermsConditionsResponse processInput(
+            TermsConditionsResponse input, RequestInfo requestInfo, Context context) throws ApiGatewayException {
+        return service.updateTermsAndConditions(
+                requestInfo.getPersonCristinId(),
+                input.termsConditionsUri(),
+                requestInfo.getCurrentCustomer()
+        );
+    }
```
truhacevkir, line 44: Nice, but since you have implemented a service for it and a new handler, we may not need a new table. Cause users-and-roles-table has already secondary index which lists all users by cristin id:
`List<UserDto> usersByCristinId = identityService.getUsersByCristinId(cristinId)`

Maybe use existing table for it? and persist it there? 
You still can use your service, but with another table. 

Simply agree terms for all users with provided cristinId?

template.yaml
```diff
@@ -698,6 +718,31 @@ Resources:
         - Key: IncludedInBackup
           Value: 'true'
 
+  NvaPersonInfoTable:
```
truhacevkir, line 721: Looks good, but not sure we need more tables.

StigNorland: a single table for the whole system is possible, and there are many good reasons for doing so....

PR #385, opened 2024-11-15
Np 47873/terms conditions

StigNorland:
I finally found the code.
one reason that it was lost, was that I did a refactoring of all the parameters in this class, and it was too many changes in one go.... still is many...


ketilaa (changes requested):
As we expect a different behaviour in `UserSelectionUponLoginHandler`, I would like to see a new test `UserSelectionUponLoginHandlerTest#shouldClearUserAttributesAndReturnRestrictedTokenWhenLatestTermsAreNotAccepted()`

ketilaa:
You are aware that @torbjokv is completing the "return token without access right claims if latest version of terms are not approved"? You are still committing code to this branch, so I think you and @torbjokv need to talk.

ketilaa:
Build is failing. Re-request review when fixed.

dynamo-client/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
     implementation libs.aws.auth
     implementation libs.aws.regions
 
+    implementation libs.dynamoDbLocal
```
ketilaa, line 14: testImplementation?

user-access-handlers/src/test/java/no/unit/nva/handlers/HandlerTest.java
```diff
@@ -59,12 +62,12 @@ private void insertRole(RoleDto role) {
                 databaseService.addRole(role);
             }
         } catch (Exception e) {
-            e.printStackTrace();
+            logger.error("Error inserting role", e);
```
ketilaa, line 65: Good. Get rid of code smells like `e.printStackTrace()`

ketilaa (approved):
Approved given that you fix gradle issue commented.

torbjokv (approved):
good stuff, lets rethink UserSelectArguments at a later date 👍

PR #389, opened 2024-11-27
refactor and add userAttributes as claims override

torbjokv:
this works with exception of "cognito:groups" that I can't find any way to override with V_2


torbjokv:
this was solved with https://github.com/BIBSYSDEV/nva-identity-service/pull/410

PR #393, opened 2024-11-29
refactor(dynamodb): use local testImplementation for local dynamodb c…

torbjokv:
…lient, upgrade dep, remove deps


user-access-testing/src/main/java/no/unit/nva/database/DatabaseTestConfig.java
```diff
@@ -3,7 +3,10 @@
 import com.amazonaws.services.dynamodbv2.local.embedded.DynamoDBEmbedded;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 
-public class DatabaseTestConfig {
+public final class DatabaseTestConfig {
+
+    private DatabaseTestConfig() {
+    }
```
StigNorland, line 9: 👍

StigNorland (approved):

PR #397, opened 2024-12-03
hotfix/template-name-fail-fix

StigNorland:


template.yaml
```diff
@@ -718,7 +719,7 @@ Resources:
         - Key: IncludedInBackup
           Value: 'true'
 
-  NvaPersonInfoTable:
+  NvaIdTypeObjectsTable:
```
torbjokv, line 722: heads up, this can delete table, but I guess its ok

template.yaml
```diff
@@ -718,7 +719,7 @@ Resources:
...
     Type: AWS::DynamoDB::Table
     UpdateReplacePolicy: Retain
     DeletionPolicy: Delete
```
torbjokv, line 725: what does nva-id-type-persistent-objects mean btw?

template.yaml
torbjokv: or leave garbage, then we need to manually tidy in every env

user-access-service/src/test/java/no/unit/nva/database/TermsAndConditionsServiceTest.java
```diff
@@ -20,6 +20,9 @@ public class TermsAndConditionsServiceTest {
     static void initialize() {
         var client = DatabaseTestConfig
                 .getEmbeddedClient();
+        var TABLE_NAME = new Environment()
```
torbjokv, line 23: this is a test, do you need to use environment here at all?

user-access-service/src/test/java/no/unit/nva/database/DynamoCrudServiceTest.java
```diff
@@ -15,18 +16,20 @@
...
-    private static DynamoCrudService<TermsConditions> termsConditionsService;
+    private static DynamoCrudService<TermsConditions> dynamoCrudService;
 
 
     @BeforeAll
     static void initialize() {
+
         var client = DatabaseTestConfig
                 .getEmbeddedClient();
+        var tableName = new Environment()
```
torbjokv, line 27: I would just hard code this name like before?

template.yaml
StigNorland: Table is new and unused, but thanks for warning me!

template.yaml
StigNorland: it is a reference to the single-table-pattern used.
`id` and `type` as a composite key access pattern.

torbjokv: Hmmm I would prefer if it have a name that reflects its content/group of content, this is in my head way to generic. if we already are changing logic id here, its a good time to change table name also.

StigNorland: so evey entity can have a host of classes assosiated with it, but only one instanse of each class.

| id   | type            | createdBy | created     | modifiedBy | modified    | ...objectfields |
|------|-----------------|-----------|-------------|------------|-------------|-----------------|
| xyz  | Person          | xyz       | 10-01-2021  | xyz        | 01-01-2022  | personinfo....  |
| xyz  | Roles           | admin     | 00-01-2021  | admin      | 01-01-2023  | RolesInfo....   |
| xyz  | UserPreferences | admin     | 00-01-2021  | xyz        | 01-10-2024  | RolesInfo....   |

torbjokv: probably do a specific table name in a later pr

user-access-service/src/main/java/no/unit/nva/database/TermsAndConditionsService.java
```diff
@@ -33,18 +33,22 @@ public TermsAndConditionsService(DynamoDbClient client, String tableName) {
...
+                                      .fetch(crudService)
+                                      .termsConditionsUri();
+                       } catch (NotFoundException e) {
+                           return null;
+                       }
+                   })
+                   .map(fetchedUri -> TermsConditionsResponse.builder()
+                                          .withTermsConditionsUri(fetchedUri)
+                                          .build())
+                   .orElse(null);
```
StigNorland, line 51: 🥇

user-access-service/src/main/java/no/unit/nva/database/TermsAndConditionsService.java
```diff
@@ -33,18 +33,22 @@ public TermsAndConditionsService(DynamoDbClient client, String tableName) {
     }
 
     public TermsConditionsResponse getTermsAndConditionsByPerson(URI cristinId) {
```
StigNorland, line 35: cristinPersonId

torbjokv (approved):

PR #404, opened 2025-01-21
ci: Replace GitHub workflows

LarsV123:


.github/workflows/cicd.yaml
LarsV123, line 1: This was disabled and unused, so I am deleting it without a replacement

PR #406, opened 2025-01-30
added last path element of viewing scope including and excluding uris…

ketilaa:
… to cognito user attributes

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1502,6 +1502,8 @@ Resources:
         - custom:impersonatedBy
         - custom:currentTerms
         - custom:acceptedTerms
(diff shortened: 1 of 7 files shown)


torbjokv (dismissed):

template.yaml
```diff
@@ -1349,6 +1349,14 @@ Resources:
           AttributeDataType: String
           Mutable: true
           Required: false
+        - Name: viewingScopeIncluded
+          AttributeDataType: String
+          Mutable: true
```
torbjokv, line 1354: nah, false maybe?

torbjokv (approved):

PR #410, opened 2025-03-11
feature(cognito): refactor, upgrade to essentials and enable access token claims

torbjokv:
also some gradle stuff


ketilaa (approved):
Trengte den refactoreringen i handleren ;)

PR #424, opened 2025-04-09
NP-48635 Add GET endpoint for channel claims

joachimjorgensen:
What is done:
- Created handler `ListAllChannelClaimsHandler`
- Added endpoint to swagger
- Renamed `channel` to `channelId`


customer-commons/src/main/java/no/unit/nva/customer/model/channelclaim/ChannelClaimDto.java
```diff
@@ -0,0 +1,18 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import no.unit.nva.commons.json.JsonSerializable;
+import nva.commons.core.JacocoGenerated;
+
+import java.net.URI;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonTypeName("ChannelClaim")
+public record ChannelClaimDto(URI channelId, ChannelConstraintDto constraint) implements JsonSerializable {
```
brinxmat, line 12: Why "channelId"? I think the "channel" was better as this is what the data represents, not "what it is" -> in our scheme "id" means "uri", so here you're saying channelUri, and the type is already clear.

customer-commons/src/main/java/no/unit/nva/customer/model/channelclaim/ChannelClaimWithClaimer.java
```diff
@@ -0,0 +1,7 @@
+package no.unit.nva.customer.model.channelclaim;
+
+import java.net.URI;
+
+public record ChannelClaimWithClaimer(ChannelClaimDto channelClaim, URI customerId, URI cristinId) {
```
brinxmat, line 5: I guess we have used "xId" elsewhere, this was a mistake, but maybe a consistent one.

customer-commons/src/main/java/no/unit/nva/customer/service/impl/DynamoDBCustomerService.java
```diff
@@ -141,6 +143,19 @@ public void createChannelClaim(UUID customerIdentifier, ChannelClaimDto channelC
         putCustomer(customerIdentifier, customer.addChannelClaim(channelClaim), false);
     }
 
+    @Override
+    public Collection<ChannelClaimWithClaimer> getChannelClaims() {
+        return getCustomers().stream()
+                   .flatMap(customer -> customer.getChannelClaims().stream()
+                                            .map(claim -> getChannelClaimWithClaimer(claim, customer)))
+                   .collect(Collectors.toList());
```
brinxmat, line 151: Are we on <java21? if not, simply toList. Fair enough to say this method returns `List<ChannelClaimWithClaimer>` since it what is being returned. Postel's law "be conservative in what you do, be liberal in what you accept from others", I guess.

truhacevkir: Here is java 17, yes :)

customer-commons/src/main/java/no/unit/nva/customer/model/channelclaim/ChannelClaimDto.java
truhacevkir: Changed back to channel :)

truhacevkir (approved):

PR #425, opened 2025-04-09
fix: Rename custom log groups

LarsV123:
Sets a specific, non-default name on all LogGroups based on the handler name.
This is necessary because a default LogGroup already exists for each handler, which we want to replace with a custom log group where we can control the retention period. Attempting to create a new LogGroup using the default name (which we previously did) is not permitted and fails during CloudFormation deployment.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -290,6 +290,8 @@ Resources:
   GetRoleHandler:
     Type: AWS::Serverless::Function
     Properties:
+      LoggingConfig:
+        LogGroup: !Ref GetRoleHandlerLogGroup
       CodeUri: user-access-handlers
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -309,12 +311,14 @@ Resources:
   GetRoleHandlerLogGroup:
     Type: AWS::Logs::LogGroup
     Properties:
-      LogGroupName: !Sub '/aws/lambda/${GetRoleHandler}'
+      LogGroupName: !Sub '/nva-identity-service/GetRoleHandler/${AWS::AccountId}'
```
brinxmat, line 314: I guess it is the intent that the name should be stable

brinxmat (approved):

PR #427, opened 2025-04-10
endpoint for fetching channel claim

truhacevkir:
Adding handler for fetching channel claim.

Now we will return following in response:

```
{
  "type": "ClaimedChannel",
  "claimedBy": {
    "type": "Customer",
    "id": "https://localhost/170e932e-ef89-4d03-899e-fabc6f2b15f2",
    "cristinId": "https://www.example.com/jbcDdKE9hpo"
  },
  "channelClaim": {
    "type": "ChannelClaim",
    "channelId": "https://host/publication-channels-path/publisher/e114f652-7467-4ccf-836a-d2ed5217acdc",
    "constraint": {
      "publishingPolicy": "Everyone",
      "editingPolicy": "Everyone",
      "scope": [
        "ChapterConferenceAbstract",
        "JournalIssue",
        "StudyProtocol"
      ]
    }
  },
  "id": "https://host/customer/channel-claim/e114f652-7467-4ccf-836a-d2ed5217acdc"
}
```
 
Pay attention to id. In next pr's I think we want to add new base path `/channel-claim` so channel-claim id makes more sense semantically.

### DIFF
--- a/customer/src/main/java/no/unit/nva/customer/get/FetchChannelClaimHandler.java
+++ b/customer/src/main/java/no/unit/nva/customer/get/FetchChannelClaimHandler.java
@@ -18,6 +18,7 @@ public class FetchChannelClaimHandler extends ApiGatewayHandler<Void, ChannelCla
 
     private static final String CLAIM_NOT_FOUND = "Channel claim with identifier %s not found";
     private static final String NOT_VALID_IDENTIFIER = "Provided channel claim identifier is not valid!";
+    private static final String IDENTIFIER_PATH = "identifier";
     private final CustomerService customerService;
 
     @JacocoGenerated
@@ -45,7 +46,7 @@ public class FetchChannelClaimHandler extends ApiGatewayHandler<Void, ChannelCla
     }
 
     private static UUID getIdentifier(RequestInfo requestInfo) throws BadRequestException {
-        return attempt(() -> requestInfo.getPathParameter("identifier"))
+        return attempt(() -> requestInfo.getPathParameter(IDENTIFIER_PATH))
                    .map(UUID::fromString)
                    .orElseThrow(failure -> new BadRequestException(NOT_VALID_IDENTIFIER));
     }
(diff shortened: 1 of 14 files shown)


truhacevkir:
I will add openapi spec and lambda function in next pr, maybe with new base path `/channel-claim` :)

customer/src/main/java/no/unit/nva/customer/get/FetchChannelClaimHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        this.customerService = customerService;
+    }
+
+    @Override
+    protected void validateRequest(Void unused, RequestInfo requestInfo, Context context) throws ApiGatewayException {
+        userIsAuthorized(requestInfo);
+    }
+
+    private static void userIsAuthorized(RequestInfo requestInfo) throws UnauthorizedException {
+        requestInfo.getCurrentCustomer();
```
truhacevkir, line 41: For now checking simply if user is authorized. Should take in use new access right once commons will be bumped in all projects.

ketilaa:
We should really stop using the term `cristin` in any names. In tghe above case, I suggest using `organizationId` instead!

ketilaa (approved):
Consider getting rid of `cristinId` in externally visible DTO's in favor of `organizationId`.

customer/src/main/java/no/unit/nva/customer/get/FetchChannelClaimHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import nva.commons.apigateway.RequestInfo;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.apigateway.exceptions.BadRequestException;
+import nva.commons.apigateway.exceptions.NotFoundException;
+import nva.commons.apigateway.exceptions.UnauthorizedException;
+import nva.commons.core.JacocoGenerated;
+
+public class FetchChannelClaimHandler extends ApiGatewayHandler<Void, ChannelClaimResponse> {
+
+    private static final String CLAIM_NOT_FOUND = "Channel claim with identifier %s not found";
```
joachimjorgensen, line 20: Should probably avoid having (malicious) user input in response? This gets stored as a log entry on a server somewhere?

joachimjorgensen: Can see now that illegal identifier throws an exception before this, so maybe it's okay.

joachimjorgensen (approved):
Minor comments, looks good!

PR #429, opened 2025-04-11
adding lambda function and openapi spec for get channel claim handler

truhacevkir:


template.yaml
```diff
@@ -1118,6 +1118,36 @@ Resources:
...
+          ID_NAMESPACE: !Sub 'https://${ApiDomain}/${CustomerDomainBasePath}'
+          API_HOST: !Ref ApiDomain
+      Policies:
+        - DynamoDBCrudPolicy:
+            TableName: !Ref NvaCustomersTable
+      Events:
+        PostChannelClaimEvent:
+          Type: Api
+          Properties:
+            Path: /channel-claims/{identifier}
```
truhacevkir, line 1141: Do it for now to deliver sprint goal and so frontend can begin with implementation. 
We should introduce new base path for channel claim. Something like /channel-claim/{identifier} (without /customer)

joachimjorgensen (approved):

PR #438, opened 2025-04-24
allowing backend client to fetch channel claim

truhacevkir:
Allowing internal client to fetch channel claim and adding missing environment variable to template.


customer/src/main/java/no/unit/nva/customer/get/FetchChannelClaimHandler.java
```diff
@@ -39,7 +39,9 @@ protected void validateRequest(Void unused, RequestInfo requestInfo, Context con
     }
 
     private static void userIsAuthorized(RequestInfo requestInfo) throws UnauthorizedException {
-        requestInfo.getCurrentCustomer();
+        if (!requestInfo.clientIsInternalBackend()) {
+            requestInfo.getCurrentCustomer();
+        }
```
truhacevkir, line 44: Backend client needs to retrieve channel ownership when validating update and publishing.

torbjokv (approved):

PR #451, opened 2025-05-28
making list customers endpoint open

truhacevkir:
This is necessary to check if contributor institution is nva customer. 

Following endpoint serves minimal customer data as cristinId, nvi institution, doi prefix and display name.


torbjokv (approved):
this is the shared data:

    private URI id;
    private URI cristinId;
    private String displayName;
    private Instant createdDate;
    private boolean active;
    private String doiPrefix;
    private boolean nviInstitution;
    
probably ok

PR #452, opened 2025-05-28
fix(UserSelectionUponLoginHandler): fix wrong name showing in gui by …

torbjokv:
…overriding idtoken

Name shows up as [%22Torbj%C3%B8rn%22] in frontend because it takes this value from idToken. New claims mapping caused value to be directly passed through to idToken and show up with bad format. To make sure frontend uses the correct value, we have to pass the value to the id token override in pre-token.

### DIFF
--- a/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
+++ b/cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
@@ -453,17 +453,25 @@ public class UserSelectionUponLoginHandler
                                  .toList();
 
         var claims = userAttributes.stream()
-                         .filter(a -> !excludedClaims.contains(a.name()) && nonNull(a.value()) && !a.value().isEmpty())
+                         .filter(attribute -> shouldBeIncludedExcept(attribute, excludedClaims))
                          .collect(Collectors.toMap(AttributeType::name, AttributeType::value));
 
         return IdTokenGeneration.builder().withClaimsToAddOrOverride(claims).withClaimsToSuppress(excludedClaims.toArray(String[]::new)).build();
+    }
+
+    private static boolean shouldBeIncludedExcept(AttributeType a, List<String> excludedClaims) {
+        return !excludedClaims.contains(a.name()) && nonNull(a.value()) && !a.value().isEmpty();
+    }
+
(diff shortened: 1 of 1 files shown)


cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -443,15 +443,20 @@ private ClaimsAndScopeOverrideDetails buildOverrideClaims(List<String> groupsToO
...
-    private IdTokenGeneration buildIdTokenGeneration() {
+    private IdTokenGeneration buildIdTokenGeneration(List<AttributeType> userAttributes) {
         var excludedClaims = Stream.concat(Arrays.stream(CLAIMS_TO_BE_INCLUDED_IN_ACCESS_TOKEN),
                                            Arrays.stream(CLAIMS_TO_BE_SUPPRESSED_FROM_PUBLIC))
-                                 .toArray(String[]::new);
-        return IdTokenGeneration.builder().withClaimsToSuppress(excludedClaims).build();
+                                 .toList();
+
+        var claims = userAttributes.stream()
+                         .filter(a -> !excludedClaims.contains(a.name()) && nonNull(a.value()) && !a.value().isEmpty())
```
brinxmat, line 456: Extract the filter condition into a named method

cognito-pre-token-generation-openid/src/main/java/no/unit/nva/cognito/UserSelectionUponLoginHandler.java
```diff
@@ -443,15 +443,20 @@ private ClaimsAndScopeOverrideDetails buildOverrideClaims(List<String> groupsToO
         return ClaimsAndScopeOverrideDetails.builder()
                    .withGroupOverrideDetails(groups)
                    .withAccessTokenGeneration(buildAccessTokenGeneration(userAttributes))
-                   .withIdTokenGeneration(buildIdTokenGeneration())
+                   .withIdTokenGeneration(buildIdTokenGeneration(userAttributes))
                    .build();
     }
 
-    private IdTokenGeneration buildIdTokenGeneration() {
+    private IdTokenGeneration buildIdTokenGeneration(List<AttributeType> userAttributes) {
```
brinxmat, line 450: Collection

